PR #241, opened 2025-06-08
fix: active campaign versions should link to gang, not campaign

### LINKED ISSUE #234: Active campaign versions should link to gang, not campaign
We have text on a non-campaign list view that says something like:

> This list has active campaign versions:
> [Example (1 — Clone Test)](http://.../campaign/2633be7f-d55a-430a-82ec-eaf4bd878071) (admin)

This should link to the campaign version of the list too.

tgvashworth:
Fixes #234

Updated the link in the list template so that active campaign versions link to the campaign version of the list (the gang) instead of the campaign itself.

This provides a better user experience by taking users directly to the campaign version of their list when they click on it.

### DIFF
--- a/gyrinx/core/templates/core/includes/list.html
+++ b/gyrinx/core/templates/core/includes/list.html
@@ -115,7 +115,7 @@
                 <ul class="mb-0 mt-1">
                     {% for clone in list.active_campaign_clones %}
                         <li>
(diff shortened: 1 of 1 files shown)


tgvashworth:
@claude Looks good but I'd actually like to like to both the list and the campaign, so you can jump to one or the other.

PR #248, opened 2025-06-08
feat: update injury system with default outcome and grouping

### LINKED ISSUE #245: Injury system tweaks
Here's a grab bag of injury system fixes we need to make:

* The injury Phase is now the "Default Outcome" as it is use to for the state we move the fighter into when this injury is applied. There's no need to rename the field, but we should update labels and help text: e.g. `The phase that this injury applies to (recovery, convalescence, permanent, or out cold).` is wrong. The options here should match the fighter states, and include a "No Change" option.
* It would be useful to have a free text "group" field on an injury that the "Select Injury" dropdown groups them into (use `group_select`) here
* The Select Injury dropdown is still showing the "Phase". Taking into account the above change around outcome for the fighter, it would be better to remove this.

tgvashworth:
Fixes #245

This PR implements the requested injury system tweaks:

- Renames injury "phase" to "default outcome" to better reflect its purpose
- Updates choices to match fighter states with a "No Change" option
- Adds a free text "group" field for organizing injuries
- Updates the injury dropdown to use grouped display
- Removes phase from the injury display string


tgvashworth:
@claude Several test files will need updating to use the new field names and choices. These weren't updated as part of this implementation but need to be addressed before merging.

tgvashworth:
@claude The tests are failing

PR #261, opened 2025-06-09
fix: prevent concurrent builds to ensure sequential deployment

tgvashworth:
- Add step to cancel any ongoing builds when a new build starts
- Add queueTtl configuration to expire builds waiting longer than 1 hour
- Add timeout configuration to prevent builds from running indefinitely
- Update notification to indicate previous builds were cancelled

This ensures that deployments always happen in the order commits were merged, preventing the issue where a slower earlier build could overwrite a newer deployment.


tgvashworth:
@claude The filter value in the gcloud command may break if $_TRIGGER_ID or $BUILD_ID contain hyphens or non-numeric characters, since they aren't quoted inside the filter expression. To ensure correct parsing, wrap substituted values in quotes (e.g., substitutions._TRIGGER_ID='$_TRIGGER_ID' AND id!='$BUILD_ID'). Hardcoding the queue TTL and timeout values makes future adjustments more error-prone. Consider defining these durations as substitutions or variables at the top of the config to centralize configuration and improve readability.

tgvashworth:
@claude Research of a local testing strategy for changes to this file, so we can confidently deploy it without breaking our pipeline

PR #268, opened 2025-06-09
feat: Compress campaign info in list view

### LINKED ISSUE #267: Compress campaign info in list view
The UI currently says "This is a campaign version of ..." and then links to the parent (cloned-from) list and the campaign.

Instead, let's move this inline with the owner and public/private info line.

For the list link, use the Copy icon. Include a tooltip that says "This list was created via cloning".

For the campaign link, use the award icon, and have a campaign status indicator: Pre and Post should be secondary badge, with In Progress as success badge.

For lists that have clones ("This list has active campaign versions"), switch to a single link with Award Icon saying "Active in Campaigns", and create a new page which shows a table of all the clones and the campaigns they are in, with campaign status (as above).

As always, use the link_with_theme tag.

tgvashworth:
This PR implements UI changes to compress campaign information in the list view.

## Changes

- Move campaign info inline with owner/public info
- Use Copy icon for parent list link with tooltip
- Use Award icon for campaign link with status badges
- Add "Active in Campaigns" link for lists with clones
- Create new page showing all campaign clones in a table
- Add link_with_theme template tag

Fixes #267

### DIFF
--- a/gyrinx/core/templates/core/includes/list.html
+++ b/gyrinx/core/templates/core/includes/list.html
@@ -37,6 +37,44 @@
                             </span>
                         {% endif %}
                     </div>
+                    {% if list.is_campaign_mode and list.original_list %}
+                        <div class="text-secondary">
+                            <i class="bi-copy"></i>
+                            {% url 'core:list' list.original_list.id as original_url %}
+                            <a href="{{ original_url }}"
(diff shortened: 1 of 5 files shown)


tgvashworth:
Claude went haywire here.

PR #283, opened 2025-06-10
fix: filter campaign actions by specific gang instead of gang owner

### LINKED ISSUE #282: Bug: Incorrect actions in the List/Gang Campaign Resources pane
The "Actions" panel on the list view is showing actions for all gangs, rather than filtered for actions linked to the _specific_ gang in view.

It should instead be filtered just for actions on the gang the user is looking at.

tgvashworth:
## Description

Fixed the bug where the "Actions" panel on the list view was showing actions for all gangs owned by the user instead of just the specific gang being viewed.

## Changes

- Changed the filter in `CampaignActionList` view to filter by specific gang ID instead of all gangs owned by the user
- Added comprehensive tests to ensure proper filtering

## Testing

Added tests to verify:
- Each gang only sees its own actions
- Multiple gangs owned by the same user don't see each other's actions
- Gangs with no actions show no actions

Fixes #282

### DIFF
--- a/gyrinx/core/tests/test_campaign.py
+++ b/gyrinx/core/tests/test_campaign.py
@@ -365,6 +365,8 @@ def test_campaign_action_list_filtering():
     action1 = CampaignAction.objects.create(
         campaign=campaign,
         user=user1,
+        owner=user1,  # Add owner field
+        list=list1,  # Associate with Gang Alpha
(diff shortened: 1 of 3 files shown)


tgvashworth:
@claude Tests are failing — take a look?

PR #297, opened 2025-06-11
feat: implement basic fighter advancements

### LINKED ISSUE #285: Basic Fighter advancements
The next big feature on our list that builds on top of the recent addition of XP support is support for advancements.

Essentially advancements allow a player to spend XP to improve the fighter.

There are several ways the player can improve the fighter and we're going to need to build support for each of them.

Every advancement has three components:

1. The XP spend
2. The modification to the fighter
3. The increase in the fighter's cost

The modifications come in a couple of flavours: adding a skill, or modifying a stat.

To make this work we need to build the advancement in userspace as a `ListFighterAdvancement` model which links to a fighter skill or stat mod.

Advancements should also _optionally_ have a link to a `CampaignAction` which will be use to track the roll they may have done for this advancement. 

The flow through this feature will be gated on an initial choice of whether or not to roll 2d6. If the player chooses to roll 2d6, we will immediately create a campaign action with the 2d6 roll and then pass the ID of that campaign action into the next form we show. The description of this action should be "Rolling for advancement to {fighter name}"

If the user does not roll 2d6, and we'll just move them to the Advancement Creation form.

That form will have an input for XP, an input for credit increase to the fighter of the advancement, and a select box with options for choosing what advancement the fighter will receive. It should also show the Action's dice roll, if created.

Once they click "Next" they will automatically be shown one of the following, based on selections on the previous screen. The advancement options are split into four types: One, improving a specific stat by name. Two, using a specific skill from the fighter's primary or secondary skill list. Three, using a random skill from the fighter's primary or secondary skill tree. And four, choosing a random skill from any skill tree. 

The options are:
- Stats — all Type One
  - Willpower
  - Intelligence
  - Leadership
  - Cool
  - Initiative
  - Movement
  - Weapon Skill
  - Ballistic Skill
  - Strength
  - Toughness
  - Wounds
  - Attacks
- Skills
  - Random Primary Skill — Type Three
  - Chosen Primary Skill  — Type Two
  - Random Secondary Skill — Type Three
  - Chosen Secondary Skill  — Type Two
  - Promote to Specialist (Same as Random Primary Skill)  — Type Three
  - Random Skill (Any Set)  — Type Four

The choices of those four options should be on a subsequent screen which, when confirmed, will create the advancement and spend the XP. And at this stage we should either update the outcome of the action we already created when we rolled the dice, or if no action was created because we didn't roll, create a new action with a description and outcome that describes what advancement the fighter received.

The description of the action if we create it at this point should be "{fighter name} spent {N} XP to advance".

Outcome of the action in either case should describe how the fighter advanced. "Gained skill" or "Improved Weapon Skill"

After that's saved, we'll show a summary screen of how the fighter advanced, what XP was spent, what XP remains, what the cost increate of the fighter is, and then a button to take them back to the list.

The Advancement should then be complete. If a dice roll happened, the advancement must link to it.

After advancement, the fighters should have:
* Lower XP, based on what the player "spent"
* Higher cost
* An update skill or improved statline

Options to edit advance and add advancements should appear on the fighter card next to the injuries. The player should be able to open a list of all the advancements, pulling in the information from the campaign action: description, outcome, how much XP it spent and what it added to the fighter cost. Within the fighter card itself we just need to count the number of advancements so it should initially say "None", then "1", "2" and so on, with the numbers in a badge. The link should say "Edit", and they should be able to add an advancement from the advancement table view.

For tracking advancement date, use the `created` field from AppBase for advancement date.

As you go through the Advancement flow, make sure to capture the parameters in the URL, or in hidden form fields. Do not create the Advancement itself until the confirm screens.

There are limits to how much a stat should be improved but we're going to leave this to the player. The dice rolling is for record keeping and utility for the player. There is no formula for how much the fighter cost should increase at the moment.

tgvashworth:
This adds support for fighter advancements in campaign mode, allowing players to spend XP to improve their fighters through stat increases or new skills.

Closes #285


tgvashworth:
@claude I've made some small tweaks. Integrate further review feedback, but think carefully about the security issue.

Make sure to show the fighter's available XP in the new `AdvancementTypeForm`, and include the missing title block.

Major change:

The "Select Skill" screen for random skills should allow the player to choose a `ContentSkillCategory` to select from (based on the `ContentFighter` `primary_skill_categories` or `secondary_skill_categories`) as relevant. So the flow should be:

* Select one of the skill advancements
* If random: allow picking of a category, then auto-select a skill
* If choice: show all skills from relevant categories

The player does not get "take-backs" on random skills

tgvashworth:
@claude Remove the "Confirm Stat Increase" screen and jump straight to the confirmation screen. Remove the "Accept this skill" screen and jump straight to the final confirmation screen (i.e. as soon as the skill is generated, we are done).

Don't use the word "Purchase" → just "Confirm Advancement". Use the symbol "¢" rather than the word credits.

CRITICAL CHANGE: Do NOT directly edit `fighter.cost_override`, instead _compute_ the cost after advancements in the fighter's `cost_int` and `cost_int_cached` methods.

The "Back to list" button is navigating back to the previous screen, rather that returning to the list.

Most forms templates for advancements are missing titles.

Have the "View Action" include a filter for the relevant list and search for "advance".

tgvashworth:
@claude Form errors are not showing up on advancement forms so, for example, errors relating to "not enough XP" are not visible.

The new advancement form is still missing a title: you should use `block head_title` rather than `block title`.

CRITICAL: Make sure that the campaign action with the advancement dice roll is done right at the start. It looks like the campaign action ID isn't being passed through correctly: the IDs don't line up, and the advancement isn't pulling through the dice roll.

CRITICAL: Make sure that advancements that don't use a roll still create a `CampaignAction` at confirmation.

tgvashworth:
@claude The forms are still missing errors when they apply to the whole form. Review the forms you have and add a snippet like:

```htmldjango
{% if form.errors %}
    <div class="alert alert-danger">
        {% if form.errors %}
            {% for field in form %}
                {% for error in field.errors %}{{ error }}{% endfor %}
            {% endfor %}
            {% for error in form.non_field_errors %}{{ error }}{% endfor %}
        {% endif %}
    </div>
{% endif %}
```

You haven't updated to `head_title` across _all_ forms you added:

- gyrinx/core/templates/core/list_fighter_advancement_confirm.html
- gyrinx/core/templates/core/list_fighter_advancement_select.html
- gyrinx/core/templates/core/list_fighter_advancement_type.html

You should run all formatters before you commit:
- ruff
- `npm run fmt`
- `djlint --profile=django --reformat --lint .`

PR #303, opened 2025-06-11
feat: Rebuild advancements

### LINKED ISSUE #302: Bug: most recent advancement commit broke the flow
The recent introduction of `sanitize_advancement_params` has broken the advancement flow.

This is because it gets the allowed params wrong. The actual code that does advancement uses the `advancement_choice` such as `stat_movement` but the recent change to sanitized parameters looks for `type` throughout as its expected param, and expects it to be `stat` or `skil`. 

This is wrong and, more concerning, the tests didn't cover this.

Review how we sanitize this data and how the form fields flow through each view to make sure that the parameters we expect are consistent and that these code paths are fully tested.

tgvashworth:
Fixes #302

The flow and data passing through the advancements was... wrong. So I rebuilt it with a bit more sanity.

This commit:
- Rebuilds the flow step by step
- Makes sure randomly selected skills don't get a "confirm" step
- Make sure campaign actions are updated

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -328,3 +328,35 @@ The script will:
 5. Create a backup of the original file before updating
 
 **Important for Claude Code**: When working on this repository, check the last date in CHANGELOG.md. If it's more than 2 days old, proactively offer to run the changelog update script to ensure the changelog stays current with recent development activity.
+
+### Fighter Advancement System
+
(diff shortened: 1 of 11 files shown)


tgvashworth:
@claude Good, but you forgot to commit and push. Try again.

PR #304, opened 2025-06-11
feat: add loading spinner to form submit buttons

### LINKED ISSUE #301: Add Simple Loading Spinner to all submit buttons.
Currently the user gets no feedback when they click on Form Submit buttons. This is a problem because our backend can sometimes be quite slow at serving requests. What I would like is a very simple piece of JavaScript that drops into every page and adds an event listener to any Form Submit buttons. So that when they're clicked it replaces the button contents with a spinner and descent Disables the button so that it can't be pressed twice.

tgvashworth:
Added a simple JavaScript enhancement that automatically shows a loading spinner on all form submit buttons when clicked. This provides immediate visual feedback to users when submitting forms, especially important when the backend response might be slow.

The implementation:
- Listens for form submit events on all forms
- Replaces button content with animated SVG spinner
- Disables buttons to prevent double-submission
- Uses the provided SVG animation for visual consistency

Fixes #301

### DIFF
--- a/gyrinx/content/templates/content/copy_selected_to.html
+++ b/gyrinx/content/templates/content/copy_selected_to.html
@@ -18,7 +18,7 @@
         {{ form }}
         <input type="hidden" name="action" value="{{ action_name }}">
         <input type="hidden" name="post" value="yes">
-        <input type="submit" value="Copy" />
+        <button type="submit">Copy</button>
         <a href="#" class="button cancel-link">{% translate "No, take me back" %}</a>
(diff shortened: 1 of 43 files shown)


tgvashworth:
@claude Update this change to migrate _all_ `<input type="submit" />` to `button` so this code works. We use it extensively.

tgvashworth:
@claude There are quite a few `<input type="submit" />` instances still in there. Take another look

PR #315, opened 2025-06-13
fix: remove campaigns group check from nav bar

### LINKED ISSUE #314: Campaign menu bar wrapped in group check
The "Campaigns" link in the nav bar is wrapped in a group check. This isn't necessary any more: we can remove it

tgvashworth:
Fixes #314

This PR removes the group check wrapper around the "Campaigns" link in the navigation bar. The link is now visible to all authenticated users instead of being restricted to the "Campaigns Alpha" group.


tgvashworth:
@claude Tests are failing.

PR #415, opened 2025-07-02
feat: Add event logging system

### LINKED ISSUE #413: Event logging
I'd like to improve the observability of user actions taken within the application in two ways. And I'd like to wrap these up in a single function call so it's really easy to augment. First I'd like a table of user actions where each action has a link to a user, a noun that it's operating on (`List`, `List Fighter`), and a verb that represents the action (`Edit`, `Archive`), as well as the obvious timestamps and unique IDs. It would also be useful to have a set of columns with useful additional context and to be defined later, and also a JSON column for further additional context that is not structured. We may show some of this to the user later, so strings should be readable. Ideally, read these from an Enum so we have a spec of all possible nouns & verbs.

Whenever one of these events is created, the full log event should be also written to the log stream as a JSON blob so that actions taken can be viewed that way. This should be an easy function call for anything in the application to make so that it's easy to pepper the whole thing with actual connections.

tgvashworth:
Closes #413

Implements a comprehensive event logging system to track user actions throughout the application.

## Changes
- Created Event model with structured fields for noun/verb tracking
- Added EventNoun and EventVerb enums for consistent action tracking
- Implemented automatic JSON logging to log stream on event save
- Created log_event() utility function for easy event creation
- Registered Event model in Django admin with appropriate permissions
- Added comprehensive tests for event logging functionality


tgvashworth:
@claude Additionally log Session ID in every event.

tgvashworth:
@claude fix the tests

tgvashworth:
@claude your last action was a mistake. the tests were failing due to an incorrect argument passed to create a list in the new tests you created. Remove the new settings_test file and instead correct the mistake introduced into the tests

tgvashworth:
@claude tests are failing:

`E       AttributeError: 'WSGIRequest' object has no attribute 'user'`

PR #526, opened 2025-07-08
fix: show equipment from both legacy and base fighters

### LINKED ISSUE #486: Issue with making Venator hunt leaders psychic
**Is your feature request related to a problem? Please describe.**
All Venator hunt leaders have the ability to upgrade to to be a non-sanctioned or sanctioned psyker for 35 creds. At present, this is implemented through two “option” items on thei equipment list. Unfortunately, once they have a gang legacy (which several do by default), their own list is completely overridden so this psyker option is no longer visible.

**Describe the solution you'd like**
Either for fighter’s gang legacies to be in addition to their underlying equipment lists, or an alternative.

**Describe alternatives you've considered**
Alternative options would be to take the equipment category “Options” (which we are using across several fighters) and making it so equipment list items that are Options are treating differently. For instance, we could rename Gear to “Gear & Options”, and make it so Options stuff is always visible, and at the top of the add/edit view?

**Additional context**

List with this issue here: https://gyrinx.app/list/eccef94e-44b7-4a30-ab49-ea2c54510d67

Image

tgvashworth:
## Summary

This PR fixes issue #486 where Venator hunt leaders with gang legacy were losing access to psyker upgrade options.

When a fighter has a legacy (e.g., Venator hunt leaders with gang legacy), the equipment list now shows items from both the legacy fighter and base fighter equipment lists. This ensures psyker upgrade options remain available.

## Changes
- Add `equipment_list_fighters` property that returns both fighters
- Update all equipment list queries to filter by both fighters
- Maintain legacy precedence for cost overrides
- Add tests to verify combined equipment lists work correctly

Fixes #486

### DIFF
--- a/gyrinx/core/models/list.py
+++ b/gyrinx/core/models/list.py
@@ -548,6 +548,17 @@ class ListFighter(AppBase):
         return self.legacy_content_fighter or self.content_fighter
 
     @cached_property
+    def equipment_list_fighters(self):
+        """
+        Return a list of fighters whose equipment lists should be considered.
+        When a legacy fighter exists, returns both legacy and base fighters
+        to allow combining their equipment lists.
+        """
+        if self.legacy_content_fighter:
+            return [self.legacy_content_fighter, self.content_fighter]
+        return [self.content_fighter]
+
+    @cached_property
     def is_stash(self):
         """
         Returns True if this fighter is a stash fighter.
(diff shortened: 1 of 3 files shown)


tgvashworth:
@claude You've added helper methods but we've got fixtures for those in conftest. Switch to using them.

PR #588, opened 2025-07-17
feat: implement flexible statline system for vehicles

### LINKED ISSUE #554: Vehicles
Holding ticket for vehicle support. We don't have anything in the Google doc on it so will be thought from scratch.

Vehicle card has:
1. Name ✅ 
2. Type ✅ 
3. Cost ✅ 
4. Characteristics: M ("), Front, Side, Rear, HP, Hnd (target), Sv (target)
5. Plus crew characteristics: BS, Ld, Cl, Wil, Int (all target) 
6. Weapons  ✅ 
7. Rules  ✅ 
8. Upgrades ❓ (if separate line, no — but looks like Psyker powers) 
9. Gear ✅ 

Immediate thoughts are that this might push us to remove the Psyker special-case and implement vehicle upgrades the same way.

## Notes

- Vehicles fighters, but are purchased as gear from the Crew equipment list
- We'll use equipment-fighter links (`ContentEquipmentFighterProfile`) to generate a vehicle fighter card
  - Don't want to make a new class because everything gets rebuilt then — gear etc still needs to be available
  - Could just hold vehicle stats in a statline class that we could later move fighter statlines to also (and remove from stash)
  - This allows for custom kinds of statlines in the content packs
- A vehicle could be in stash — again, using equipment-fighter link

# Implementation Plan

## Statline

- Build out a `ContentStatlineType` and `ContentStatline` which can be linked to a ContentFighter
- The `ContentStatlineType` should have a name (e.g. `Vehicle`), and set of stats with full name and short name:
  - Movement (M")
  - Toughness: Front (Fr), Side (Sd), Rear (Rr)
  - Hull Points (HP)
  - Handling (Hnd+) and Save (Sv+) — both "target" i.e. `4+`
- The `ContentStatline` should be linked to the `ContentFighter`
- Update the `statline()` methods of the `ContentFighter` to look for a `ContentStatline` and to use this preferentially over the built-in stats
- Test this via `statline` and `statline_display` (and it's probably time we typed these methods)
- Update the `fighter_card_content_inner.html` template to:
  - Generate the table headings from the statline display (using the `ContentStatlineType`)
  - Calculate the `colspan` values for the rest of the template so that they display correctly with 7 (and, later, 11) stats

## Upgrades

- We can do this with special categories that use the `Visible only if in equipment list` setting

## UI

- Buying vehicle then crew:
  - We could offer a `+ Add Vehicle` button which opens a special form that selects the vehicle, and the asks for it to be crewed, over two steps
- Later, we can address "merging" the two cards into something useful. In particular, the statlines will be a bit faffy: this will likely need to be done within the `ListFighter#statline` method.

tgvashworth:
Implements the "Statline" part of issue #554 for vehicle support.

## Changes

- Added ContentStatlineType model to define statline types (Fighter, Vehicle, etc)
- Added ContentStatlineTypeStat model to define individual stats with display properties
- Added ContentStatline model to store custom stat values for fighters
- Updated ContentFighter.statline() to check for custom statlines first
- Updated fighter card template to use dynamic headers based on statline
- Calculated colspan values dynamically using widthratio template tag

This allows for flexible statlines with different numbers and types of stats, enabling vehicle support with their unique stat requirements (M, Fr, Sd, Rr, HP, etc).

The implementation is backwards compatible - existing fighters continue to use the hardcoded 12-stat system until they have a custom statline assigned.

Closes #554 (Statline implementation)


tgvashworth:
@claude Add admin registration for the new models, with inlines of the `ContentStatlineTypeStat` under each `ContentStatlineType`

tgvashworth:
@claude Let's generate the field_name on first save, and make it read-only in the admin view. Clean, lowercase and snake_case the fully name.

tgvashworth:
@claude Instead of the `AdminTextareaWidget` for `ContentStatlineAdmin`, let's have a `ContentStatlineStat` model that links a `ContentStatline` and `ContentStatlineTypeStat`.

Then let's inline the `ContentStatlineStat` type within the admin of `ContentStatlineAdmin`. The validation on save of the `ContentStatline` should make sure that there is a value for each `ContentStatlineTypeStat` relevant to the selected `ContentStatlineType`

tgvashworth:
@claude Vehicles should not have skills or house additional gear, let's add a flags to ContentFighter: "hide_skills" and "hide_house_restricted_gear" and then, in the fighter card (`fighter_card_content_inner.html`) don't show the skilline or `has_house_additional_gear` sections.

PR #596, opened 2025-07-18
feat: add individual campaign asset deletion

### LINKED ISSUE #590: Need to be able to delete specific campaign assets
You can delete an entire asset category, but not individual assets within that category. Need to be able to do this in case of accidental repetition etc.

tgvashworth:
Fixes #590

## Summary
Added the ability to delete individual campaign assets, following the pattern of asset type deletion.

## Changes
- Added Remove button to each asset in the campaign assets view
- Created confirmation page before deletion
- Only campaign owner can delete assets
- Cannot delete assets from archived campaigns
- Event logging for asset deletion

## Implementation
The Remove button uses `btn-outline-danger` styling with a trash icon, as requested. The confirmation step shows which list currently holds the asset (if any).

### DIFF
--- a/gyrinx/core/templates/core/campaign/campaign_asset_remove.html
+++ b/gyrinx/core/templates/core/campaign/campaign_asset_remove.html
@@ -1,0 +1,35 @@
+{% extends "core/layouts/base.html" %}
+{% load allauth custom_tags %}
+{% block head_title %}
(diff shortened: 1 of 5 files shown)


tgvashworth:
@claude create a campaign action when the asset is deleted.

PR #599, opened 2025-07-19
fix: invert list fighter sort order for linked vehicles

### LINKED ISSUE #593: Invert list fighter sort order for linked vehicles
Currently, linked fighers are sorted by the ListFighterManager to appear _after_ their owner via "-after" suffix. Instead, they should sort _before_ their linked owner.

tgvashworth:
Linked fighters now appear before their owner instead of after.

Changed the sort key for linked fighters from appending "-after" to prepending a space character, which sorts before alphanumeric characters.

Fixes #593

### DIFF
--- a/gyrinx/core/models/list.py
+++ b/gyrinx/core/models/list.py
@@ -365,9 +365,7 @@ class ListFighterManager(models.Manager):
                     # Linked fighters should be sorted next to their parent
                     When(
                         _is_linked=True,
-                        then=Concat(
-                            "linked_fighter__list_fighter__name", Value("-after")
-                        ),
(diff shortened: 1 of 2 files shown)


tgvashworth:
@claude This is incorrect. Instead, the sort should work as follows:

- Fighter linked to Beast should sort: Fighter, Beast
- Fighter linked to Vehicle should sort: Vehicle, Fighter
- Fighter linked to Vehicle, where Fighter is also linked to Beast, should sort: Vehicle, Fighter, Fighter's Beast
- Fighter linked to Vehicle, where Vehicle is also linked to Beast, should sort: Vehicle, Fighter, Vehicles's Beast

Ideally this implementation would work a bit better so that we could be very confident in the sort output order.

PR #621, opened 2025-07-21
fix: widen left column on fighter cards to fit "Advancements"

### LINKED ISSUE #611: May need to slightly widen left column on fighter cards to fit “advancements”
“Advancements” isn’t flowing very nicely on standard iPhone view:

Image

Possibly we need to widen that column slightly? I doubt we’ll ever have a longer single word in that space so we can set the max to fix for advancements specficially.

tgvashworth:
This PR updates the fighter card templates to widen the left column from 25% to 33%, providing more space for row headers. This prevents text overflow for longer words like "Advancements" on mobile views.

Fixes #611


tgvashworth:
@claude The 75 ratios now need to be 66, otherwise we get overflow.

PR #646, opened 2025-07-22
feat: Add robots.txt to disallow crawling of /admin/

### LINKED ISSUE #645: Add appropriate robots.txt
We see 500s on /robots.txt (which is a separate bug). But this would be a useful file to have. Let's set one up with no crawling of /admin for all bots

tgvashworth:
Fixes #645

## Summary

This PR adds a robots.txt file that disallows all bots from crawling the /admin/ directory.

## Changes

- Added a `robots_txt` view in `gyrinx/pages/views.py` that serves plain text content
- Added URL pattern for `/robots.txt` in main `urls.py`
- Added test to verify robots.txt functionality

## Testing

Added a test in `test_error_handlers.py` that verifies:
- The endpoint returns 200 status
- Content-Type is `text/plain`
- Content includes the expected directives

## Generated with [Claude Code](https://claude.ai/code)


tgvashworth:
@claude I'd prefer to have a robots.txt file to edit, although we can still do it this way with a view.

PR #671, opened 2025-07-24
fix: assignment deletion cost cache recalculation

### LINKED ISSUE #648: Assignment deletion are not causing cost cache recalculation
We use signals extensively to ensure list costs are cached correctly. These typically run on save of new objects.

It looks like we have a bug where deleted assignments (e.g. removing a weapon) does _not_ cause this to update.

Action: Review the signals we use to cause list cost to be recalculated to ensure deletion is taken into account.

tgvashworth:
Changed ListFighterEquipmentAssignment signal from pre_delete to post_delete
to ensure cost cache is updated after the assignment is actually deleted.
This fixes the issue where the cached cost included the deleted assignment.

Added comprehensive tests to verify deletion correctly updates both regular
and cached cost calculations.

Fixes #648

### DIFF
--- a/gyrinx/core/models/list.py
+++ b/gyrinx/core/models/list.py
@@ -2116,7 +2116,7 @@ def delete_related_objects_post_delete(sender, instance, **kwargs):
 
 
 @receiver(
-    [pre_delete, post_save],
+    [post_delete, post_save],
     sender=ListFighterEquipmentAssignment,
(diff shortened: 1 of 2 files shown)


tgvashworth:
@claude I agree with the change but we need to extend the tests: the list cost cache is used in other places, so it's worth us checking that _immediately after_ the delete, without `refetch_from_db`, the list cost cache has updated with the correct value.

PR #681, opened 2025-07-24
feat: Allow XP and advancements outside of campaigns

### LINKED ISSUE #680: Allow XP and advancements outside of campaigns
Currently managing XP and advancements is limited to lists in campaign mode (`lst.status = List.CAMPAIGN_MODE`). However, users are finding this limiting.

We should open these two feature up to lists in any status, even if they are in `List.LIST_BUILDING` status.

When shipping this:
- Campaign actions should still be created if the list status is `CAMPAIGN_MODE`
- XP and advancements should be copied when fighers are cloned
- We should hide the "The roll will immediately be added to the campaign action log." message if not in a campaign

tgvashworth:
Fixes #680

## Summary
- Removed campaign mode restriction from XP editing and advancement flows
- Campaign actions still created when in campaign mode
- Updated UI to conditionally show campaign action message
- Fixed fighter clone to copy XP and advancements
- Updated tests to reflect new behavior

### DIFF
--- a/gyrinx/core/models/list.py
+++ b/gyrinx/core/models/list.py
@@ -1289,6 +1289,8 @@ class ListFighter(AppBase):
             "cool_override": self.cool_override,
             "willpower_override": self.willpower_override,
             "intelligence_override": self.intelligence_override,
+            "xp_current": self.xp_current,
+            "xp_total": self.xp_total,
             **kwargs,
         }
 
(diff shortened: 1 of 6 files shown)


tgvashworth:
@claude Disable the "Roll D6" button if not in campaign mode, with a tooltip saying "Only available when the gang is active in a campaign"

PR #717, opened 2025-07-31
fix: allow deselection of single-select gang attributes

### LINKED ISSUE #709: Allow deselection of single-select gang attributes
Right now it’s not possible to deselect a single-select gang attribute once it has been saved. We need to enable deselect then save.

tgvashworth:
Fixes #709

## Summary

This PR adds a "None" option to single-select gang attributes, allowing users to deselect previously selected values.

## Changes

- Added `empty_label="None"` to the `ModelChoiceField` for single-select attributes in `ListAttributeForm`
- This provides a UI way to clear single-select attribute assignments

## Test Plan

- All existing tests pass, including `test_campaign_action_clear_values` which specifically tests deselection behavior
- Manual testing: Create a list, select a single-select attribute value, save, then select "None" and save again

### DIFF
--- a/gyrinx/core/forms/attribute.py
+++ b/gyrinx/core/forms/attribute.py
@@ -30,13 +30,18 @@ class ListAttributeForm(forms.Form):
 
         if self.attribute.is_single_select:
             # Single select - use radio buttons
+            # Note: empty_label doesn't work with RadioSelect, so we manually add None option
             self.fields["values"] = forms.ModelChoiceField(
                 queryset=values,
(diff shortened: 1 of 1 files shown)


tgvashworth:
@claude This doesn't work. From the documentation: "Note that no empty choice is created (regardless of the value of empty_label) if a ModelChoiceField is required and has a default initial value, or a widget is set to RadioSelect and the blank argument is False."

So we need to add a "None" option manually

PR #720, opened 2025-07-31
fix: remove duplicate query string entries from hidden fields

### LINKED ISSUE #715: Duplicate query string entries from hidden fields
When we use hidden fields to provide default values for a checkbox switch (e.g. filter=all vs filter=equipment-list) we get duplicates. Ideally we'd remove the hidden fields and use the non-existence of the query param as the "off" state.

This happens everywhere we use the filter pattern: gear assignment, skills, lists, campaigns.

tgvashworth:
Fixes #715

## Summary
- Remove hidden input fields that were causing duplicate query parameters
- Update checkbox logic to properly handle the "all" state
- Changed from checking if filter is not set or equals "equipment-list" to checking if filter != "all"

## Test plan
- [x] Visit gear assignment page and toggle "Only Equipment List" filter
- [x] Visit skills page and toggle "Primary/Secondary Only" filter
- [x] Visit weapon accessories page and toggle filter
- [x] Verify no duplicate query parameters in URL
- [ ] Verify filter=all links work correctly

### DIFF
--- a/gyrinx/core/static/core/js/index.js
+++ b/gyrinx/core/static/core/js/index.js
@@ -367,11 +367,15 @@ document.addEventListener("DOMContentLoaded", () => {
             // and disable it when the checkbox is checked
             const form = checkbox.form || checkbox.closest("form");
             if (form) {
-                // The input may not be within (DOM Child) of the form, so we need to use document.querySelector
-                // to find it by name and form ID.
-                const hiddenInput = document.querySelector(
-                    `input[type="hidden"][name="${checkbox.name}"][form="${form.id}"]`,
-                );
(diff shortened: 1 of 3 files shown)


tgvashworth:
@claude This doesn't work:

1. When the page loads, the filter is not set, so the checkbox is checked (correct)
2. When unticking, the page loads again with no filter value (incorrect — should be `all`)
3. As a result, the box returns to ticked state (incorrect)

We need the filter value to be "all" when the box is unchecked.

Let's reset back to before this change but, before submitting the form, disable the hidden input based on the checked condition of the associated checkbox.

PR #761, opened 2025-08-03
fix: remove unsafe filter from help_text in form field template

### LINKED ISSUE #757: Remove unsafe filter from help_text in form field template
The custom Django form field template uses the `|safe` filter on help_text, which could lead to XSS vulnerabilities if help_text ever contains user-controlled content.

## Examples

In `gyrinx/templates/django/forms/field.html:11`:
```html
<div class="helptext form-text"
     {% if field.auto_id %}id="{{ field.auto_id }}_helptext"{% endif %}>{{ field.help_text|safe }}</div>
```

This template is used throughout the application, including in the fighter state edit form.

## Suggested Improvement
Remove the `|safe` filter and let Django's auto-escaping handle the help text:
```html
<div class="helptext form-text"
     {% if field.auto_id %}id="{{ field.auto_id }}_helptext"{% endif %}>{{ field.help_text }}</div>
```

If certain help texts need HTML formatting, they should use `mark_safe()` explicitly in the form definition where the developer has full control.

## Impact
- **Area**: All forms across the application
- **Severity**: High (Security vulnerability)
- **Effort**: Small

## Files Affected
- gyrinx/templates/django/forms/field.html

---
*This issue was automatically generated by Claude Code Improvement workflow*

tgvashworth:
Removed the |safe filter from field.help_text to prevent potential XSS vulnerabilities.
Django's auto-escaping will now handle the help text safely by default.
If HTML formatting is needed in specific help texts, they should use mark_safe()
explicitly in the form definition.

Fixes #757

### DIFF
--- a/gyrinx/core/forms/__init__.py
+++ b/gyrinx/core/forms/__init__.py
@@ -2,6 +2,7 @@ from allauth.account.forms import LoginForm, ResetPasswordForm, SignupForm
 from django import forms
 from django.contrib.auth import get_user_model
 from django.core import validators
+from django.utils.safestring import mark_safe
 from django.utils.translation import gettext_lazy as _
 from django_recaptcha.fields import ReCaptchaField, ReCaptchaV3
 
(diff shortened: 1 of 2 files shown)


tgvashworth:
@claude This will break our `tos_agreement` `help_text` which is used on the signup form — there is some HTML in there which needs to be rendered. Fix that

PR #815, opened 2025-08-06
fix: Show 'None' for empty vehicle upgrade slots on fighter cards

### LINKED ISSUE #807: Vehicle upgrades do not show "None"
When empty vehicle upgrade slots on the fighter card do not show the secondary text italic none to the non-owning user like they do in other parts of the fighter card.

tgvashworth:
When equipment (including vehicles) has available upgrades but none are selected, non-owning users now see 'None' in italics, consistent with other empty sections on fighter cards.

Fixes #807


tgvashworth:
Incorrect fix

PR #846, opened 2025-08-07
feat: Implement Equipment List Expansion system

### LINKED ISSUE #693: Equipment list expansion based on gang traits
In many instances, certain traits of a gang will mean certain fighters within that gang gain certain equipment on their lists. This is especially prevalent for Outcasts, but has a number of other applications, such as:

— As mentioned, Outcasts gangs get different equipment lists depending upon what affiliation they take
— Malstrain leaders and champions gaining tyramite access
— Chaos corrupted leaders and champions gaining chaos familiar access
— Post-Apocrypha Necromunda, ALL vehicles in house gangs have slightly expanded equipment lists
— GSC corrupted gangs get psychic familiar access

etc etc

With the system we have at the moment, we need to be able to link these relations based on specific gang attributes, and also based upon what the House type is.

I would envision something like “Equipment list expansion”, in which the following things can be set:

— Which specific attribute/house the expanded list is linked to (so it will permanently be in effect for certain houses)
— All the items that are gained as part of the expansion
— Per item, what type of fighter within the gang’s list benefits from the expansion

Being able to do all the items on one page would be useful from an input speed perspective.

tgvashworth:
Adds support for expanding equipment lists based on gang attributes, houses, and fighter categories.

## Features
- Polymorphic rule system with AND logic for conditions
- Rule types: by attribute, by house, by fighter category
- Cost overrides per expansion item
- Utility methods for getting applicable expansions
- Comprehensive test coverage for all scenarios
- Full Django admin interface

Closes #693


gyrinx/content/models.py
```diff
@@ -522,6 +522,54 @@ def with_cost_for_fighter(
...
+        Annotates the queryset with fighter-specific cost overrides,
+        including those from equipment list expansions.
+        """
+
+        # Filter to only expansions that apply
+        expansion_ids = []
+        for expansion in ContentEquipmentListExpansion.objects.prefetch_related(
+            "rules"
+        ).all():
+            if expansion.applies_to(list_obj, content_fighter):
```
tgvashworth, line 538: This should be `applies_to(rule_inputs)` where `rule_inputs` is a dataclass with option properties like `list_obj: Optional[List]`. This is so we can evolve and grow this interface over time. `ExpansionRuleInputs` should be passed into `applies_to` not generated within it.

gyrinx/content/models.py
```diff
@@ -3125,3 +3181,281 @@ class Meta:
...
+        choices=FighterCategoryChoices.choices,
+        help_text="Fighter categories that must match",
+    )
+
+    def match(self, rule_inputs) -> bool:
+        """Check if the fighter is one of the required categories."""
+        fighter = rule_inputs.fighter
+        if not fighter:
+            return False
+        return fighter.category in self.fighter_categories
```
tgvashworth, line 3449: It's critical that this fighter is a ListFighter, so we can use `get_category` method which allows the `category_override` to work.

gyrinx/content/models.py
```diff
@@ -1407,6 +1455,14 @@ def with_cost_for_fighter(
         )
 
 
+@dataclass
+class ExpansionRuleInputs:
+    """Inputs for evaluating equipment list expansion rules."""
+
+    list: object  # The List (gang) to evaluate - using object to avoid circular import
+    fighter: "ContentFighter" = None  # The fighter to evaluate (optional)
```
tgvashworth, line 1463: This needs to be `List` and `ListFighter`.

Move this somewhere that avoids the circular import, and fix up places that use it to ensure ListFighter and List are used.

tgvashworth (commented):
@claude Act on the review feedback from copilot and me

PR #893, opened 2025-08-13
perf: optimize fighter cost calculations with subqueries and aggregation

tgvashworth:
- Add annotated_house_cost_override subquery to avoid N+1 queries for house cost overrides
- Add annotated_advancement_total_cost with SUM aggregation to calculate advancement costs in single query
- Cache banner lookup in tests to prevent unnecessary queries
- Update performance template to show base cost, advancement cost, and cost override status
- Reduce overall query count for list performance view


tgvashworth:
@claude Verify and then act on copilot feedback

Issue #919, opened 2025-08-21
Improve list print layout


tgvashworth:
The current print view has several issues, where cards overrun the page and the sizing/spacing is wrong.

There's some AI-generated research on this topic which may or may not be helpful:
- [CSS Grid Print Layouts: Overcoming the fundamental browser limitation](https://claude.ai/public/artifacts/10673870-aa90-4e42-89bb-4df046d3485c)
- [Printing CSS Grid Cards](https://chatgpt.com/share/68a77bc9-3f40-800e-a795-b050a975b310)

The overall answer seems to be that modern browsers are/were buggy at the intersection of CSS grid and the new print stuff.

Our _ideal_ dreamland output would be:

1. Looks reasonably good
2. Fit most lists onto an A4 page (9-12 fighters)
3. No card _ever_ runs over the page
4. Customisable font-size and grid span (3/4/6/12 cols per card) for the user
5. Cards can be re-ordered in some way

Realistically, for a first pass, I think we should focus on 1, 2, and 3.

tgvashworth:
[Squats Basic List _ Ironhead Squats (HotA) _ Gyrinx.pdf](https://github.com/user-attachments/files/21925881/Squats.Basic.List._.Ironhead.Squats.HotA._.Gyrinx.pdf)

Here's a reference print at 60% scale as a baseline. It's pretty shit.

tgvashworth:
This is done!

PR #923, opened 2025-08-24
feat: implement separate print & screen stylesheets

tgvashworth:
- Split styles.scss into screen.scss and print.scss with different font sizes
- Adjust grid layouts for better print output with more columns per row
- Update package.json to compile both stylesheet variants
- Remove QR code card wrapper for cleaner print appearance

Address #919

### DIFF
--- a/gyrinx/core/static/core/scss/print.scss
+++ b/gyrinx/core/static/core/scss/print.scss
@@ -1,0 +1,7 @@
+$font-size-base: 1rem;
+
+@import "./styles";
+
+body {
+    zoom: 50%;
+}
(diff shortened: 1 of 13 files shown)


tgvashworth:
Merging because the proposed CSS alternative doesn't have the same effect, and it _does_ work in firefox:

[Squats Basic List _ Ironhead Squats (HotA) _ Gyrinx _ firefox.pdf](https://github.com/user-attachments/files/21956081/Squats.Basic.List._.Ironhead.Squats.HotA._.Gyrinx._.firefox.pdf)
[Squats Basic List _ Ironhead Squats (HotA) _ Gyrinx _ grid fix.pdf](https://github.com/user-attachments/files/21956082/Squats.Basic.List._.Ironhead.Squats.HotA._.Gyrinx._.grid.fix.pdf)

PR #956, opened 2025-09-07
Closes #806 - Use radio buttons for category filter in skill selection

### LINKED ISSUE #806: Adds skills screen buttons should be radio style
The Primary/Secondary Only button and Show Restricted button on Add Skills should have some kind of radio-style interaction. Seems like you wouldn’t want to have both on at the same time.

fiveeightsix:
Replaces "Primary/Secondary Only" and "Show Restricted" switched with single radio group, with the following options:
- "Primary/secondary only"
- "All"
- "All + restricted",
better representing the filter behaviour.

i.e.
image

and
image


fiveeightsix:
@tgvashworth it occurs to me having submitted this that the filter might make more sense positioned under the "Skill categories" subhead, and above the category list? It feels more concerned with the categories than with the individual skills, and it would give the layout more room. What do you think?

tgvashworth:
@fiveeightsix Yeah, that sounds like a good idea. You could also try out collapsing those to a dropdown menu on mobile with a "Jump to:" label, or similar.

Issue #964, opened 2025-09-08
Implement ContentAdvancement


tgvashworth:
Based on description in #605.

- Introduce `ContentAdvancementEquipment` _just_ for the cases where equipment is being selected
- These should be linked to `ListFighterAdvancement` if selected (new field required here)
- Find these available advancements in a similar way to the `ContentStat`-driven advancements we now have — i.e. make them choices on the `AdvancementTypeForm`
- Allow `ContentAdvancementEquipment` to have either or both `random`/`chosen` flags (we'll then generate the options)
- Allow content admin to give the advancement a name (e.g. `Legendary Name`) which is then prefixed with `Random` or `Chosen` as appropriate
- The form value would be `equip_` prefixed, and then generated from the ID of the advancement (so that renames are possible) e.g. `equip_random_f300e5f0-25f7-4b73-ae0f-e1eefbab3998` and/or `equip_chosen_f300e5f0-25f7-4b73-ae0f-e1eefbab3998`

tgvashworth:
This is launched

PR #972, opened 2025-09-12
feat: Implement equipment-based advancements

tgvashworth:
- Implements #964 - Equipment-based advancements with duplicate upgrade detection
- Allows fighters to gain equipment through XP expenditure
- Supports both chosen (select from list) and random equipment options
- Prevents fighters from gaining duplicate equipment upgrades through advancements

## Core Changes

- Added ContentAdvancementEquipment model with equipment assignments
- Extended ListFighterAdvancement with equipment assignment support
- Equipment advancements appear in advancement type selection form
- Chosen equipment: Shows filtered list of available equipment
- Random equipment: Goes directly to confirmation with automatic selection


tgvashworth:
@claude We don't need backwards compat: you can now additionally remove support for the older direct-equipment-linked advancement system.

PR #979, opened 2025-09-14
Increase relative fighter stat block font size on print view, and related tweaks. 

fiveeightsix:
Resolves 936.

- Font size in the fighter card stat block increased, making it easier to read (print view only).
- More space given for content in fighter detail block at expense of row label, to fit more items in (print view only).
- Removed gap between item and comma in "Rules" list.

Before:
image

After:
image

Did not change table padding, as this is already as low as is possible without customising bootstrap styles further.

### DIFF
--- a/gyrinx/core/templates/core/includes/fighter_card_content_inner.html
+++ b/gyrinx/core/templates/core/includes/fighter_card_content_inner.html
@@ -10,19 +10,19 @@
 {% endcomment %}
 {% comment %} Stats {% endcomment %}
 <table class="table table-sm table-borderless table-fixed mb-0">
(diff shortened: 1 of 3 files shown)


tgvashworth:
PDF here: [Squats Basic List _ Ironhead Squats (HotA) _ Gyrinx - font size.pdf](https://github.com/user-attachments/files/22648208/Squats.Basic.List._.Ironhead.Squats.HotA._.Gyrinx.-.font.size.pdf)

Happy to ship

tgvashworth:
@claude Undo the indentation changes on this branch. We can look at those later.

PR #1007, opened 2025-10-02
feat: add cost filter to gear/weapon screens

### LINKED ISSUE #997: Cost filter on gear/weapon screens
Would be useful to be able to filter by cost on gear/weapon screens (slider with manual entry of limit would be useful). Don’t think we need to support ranges/minimum costs.

tgvashworth:
Add maximum cost filter to equipment selection screens (both weapons and gear):
- Add Cost dropdown in fighter_gear_filter.html with slider (0-500, step 5) and number input
- Use data-gy-sync pattern to sync slider and input values
- Implement backend filtering on cost_for_fighter field in list.py
- Preserve mc parameter in query strings when adding equipment
- Filter applies after availability filtering and respects can_buy_any houses

Closes #997

### DIFF
--- a/gyrinx/core/static/core/js/index.js
+++ b/gyrinx/core/static/core/js/index.js
@@ -210,13 +210,16 @@ document.querySelectorAll("[data-gy-sync]").forEach((element) => {
 
     if (targets.length === 0) return;
 
-    element.addEventListener("change", (event) => {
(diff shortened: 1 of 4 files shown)


tgvashworth:
@claude This doesn't actually work. Write a test to verify the behaviour and fix the issue.

PR #1042, opened 2025-10-19
Add autocomplete fields to content admin screens

### LINKED ISSUE #856: Use autocomplete_fields more widely in content admin
The `autocomplete_fields` tool is very useful for content admins. We should sprinkle it more liberally through admin.

Suggestions:

* Content Fighter searches
* Equipment
* Modifiers (if possible while also allowing multi-select)

fiveeightsix:
Closes #856. Adds autocomplete widgets for fighter, fighter type, equipment, psychic power disciplines, etc. in the content admin interface, where possible. Where the fields previously used grouped drop-downs, this has been removed to add the autocomplete.

### DIFF
--- a/gyrinx/content/tests/test_equipment_upgrade_overrides.py
+++ b/gyrinx/content/tests/test_equipment_upgrade_overrides.py
@@ -57,7 +57,7 @@ def test_content_fighter_equipment_list_upgrade_creation():
 
     assert override.cost_int() == 30
     assert override.cost_display() == "30¢"
-    assert str(override) == "Archaeotek Upgrades – Suspensor (30)"
+    assert str(override) == "Archaeotek Upgrades – Suspensor (Servo-arm) (30)"
 
 
 @pytest.mark.django_db


tgvashworth:
@fiveeightsix Nice, thanks. I had thought we'd only do this for content fighter and equipment lookups, because there are so many of those. Does content admin feel better/easier with these changes extending outside of just those two types?

I should be able to give this a proper test this evening so will have thoughts myself

fiveeightsix:
@tgvashworth I think so? Admittedly I've not used the admin interface much. I was working on the premise of everything that felt I had to hunt for an option got an autocomplete, but I may have got carried away so happy to back any specific changes out if needed :)

gyrinx/content/admin.py
```diff
@@ -387,53 +387,36 @@ def __init__(self, *args, **kwargs):
...
 
 
-class ContentFighterEquipmentListUpgradeAdminForm(forms.ModelForm):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-
-        group_select(
-            self, "fighter", key=lambda x: x.house.name if x.house else "No House"
-        )
-        group_select(self, "upgrade", key=lambda x: x.equipment.name)
```
tgvashworth: I thought this one needed to stay because the `x.equipment.name` needed to differentiate between similarly-name... but in testing this, I think there was a bug here and I can't see the equipment name being used for grouping 😱 

So I suggest we make this better by updating the implementation of `__str__` on `ContentEquipmentUpgrade` to..

```python
def __str__(self):
  return f"{self.equipment.upgrade_stack_name_display} – {self.name} ({self.equipment.name})"
```

tgvashworth (changes requested):
Looks good to me apart from one small improvement we can make.

gyrinx/content/admin.py
fiveeightsix: OK, made the change to `ContentEquipmentUpgrade`, and the autocomplete on line 413 below. Does this look right?:

image

tgvashworth (approved):

PR #1065, opened 2025-11-08
fix: use cost_for_house for vehicles in stash

### LINKED ISSUE #1063: We aren't using cost_for_house for vehicles (all equipment?) in stash
This leads to a bug where, for example, a squats svenotar scout trike looks free when it's in the stash

This could should be reproduced with:
- A piece of ContentEquipment with no intrinsic cost
- A `ContentEquipmentFighterProfile` between the equipment and a (vehicle) ContentFighter
- A `ContentFighterHouseOverride` between the house and the ContentFighter
- Have the equipment assigned to stash fighter in the list

Expected:
* Stash fighter cost_int includes the vehicle cost, matching the `ContentFighterHouseOverride`

Actual:
* Some other cost (likely the ContentEquipment zero cost) is used, so the stash has 0 cost

tgvashworth:
Fixes #1063

When equipment with ContentEquipmentFighterProfile (vehicles/exotic beasts) is assigned to stash, the cost calculation now uses the child fighter's cost_for_house() method to properly apply ContentFighterHouseOverride costs.

Previously, vehicles in stash showed 0 cost (from ContentEquipment.cost) instead of the house-specific cost from ContentFighterHouseOverride.

## Changes
- Modified `_equipment_cost_with_override()` in `gyrinx/core/models/list.py`
- Added test `test_vehicle_stash_cost_bug.py` to verify the fix

### DIFF
--- a/gyrinx/core/tests/test_vehicle_stash_cost_bug.py
+++ b/gyrinx/core/tests/test_vehicle_stash_cost_bug.py
@@ -2,22 +2,27 @@
 
 import pytest
 from gyrinx.content.models import (
-    ContentEquipment,
     ContentEquipmentFighterProfile,
-    ContentFighter,
     ContentFighterHouseOverride,
-    ContentHouse,
 )
-from gyrinx.core.models import List, ListFighter, ListFighterEquipmentAssignment
+from gyrinx.core.models import ListFighterEquipmentAssignment
+from gyrinx.models import FighterCategoryChoices
 
 
 @pytest.mark.django_db
-def test_vehicle_in_stash_uses_house_override_cost(user):
-    house = ContentHouse.objects.create(name="Squats", short_name="SQT")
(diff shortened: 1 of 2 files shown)


gyrinx/core/tests/test_vehicle_stash_cost_bug.py
```diff
@@ -0,0 +1,77 @@
...
+    ContentFighter,
+    ContentFighterHouseOverride,
+    ContentHouse,
+)
+from gyrinx.core.models import List, ListFighter, ListFighterEquipmentAssignment
+
+
+@pytest.mark.django_db
+def test_vehicle_in_stash_uses_house_override_cost(user):
+    house = ContentHouse.objects.create(name="Squats", short_name="SQT")
```
tgvashworth: `short_name` is not a thing.

Use `conftest.py` for fixtures.

PR #1090, opened 2025-11-14
Change campaign resources panel from lists to table

### LINKED ISSUE #986: Change campaign resources panel from lists to table
For campaigns with lots of resources, the different categories can end up taking a lot of vertical space. User (Bainey) has suggested we instead implement a table system so each gang appears once with their resources across columns — which might also allow for better sorting across different metrics. This seems to make sense.

Illustrative screenshot here: 

Image

tgvashworth:
## Summary

Converts the campaign resources panel to display gangs as rows with resource types as columns, instead of grouping by resource type.

- Reorganize table structure: gangs as rows, resources as columns
- Remove badge styling from amounts (display as plain text)
- Use icon-only edit links (no "Modify" text)
- Add table-responsive wrapper for mobile support

Addresses feedback from user to reduce vertical space and improve sorting capabilities across different metrics.

Fixes #986

### DIFF
--- a/gyrinx/core/templates/core/campaign/campaign.html
+++ b/gyrinx/core/templates/core/campaign/campaign.html
@@ -228,24 +228,24 @@
                         <i class="bi-coin"></i> View Resources
                     </a>
                 </div>
-                <div class="card-body p-0">
+                <div class="card-body p-0 px-2 px-sm-3">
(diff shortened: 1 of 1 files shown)


tgvashworth:
@claude Implement copilot suggestions

PR #1107, opened 2025-11-17
Add defensive fix for missing campaign list resources

tgvashworth:
- Ensure all lists have resources for all types when viewing campaign detail
- Change resource allocation to use get_or_create to prevent duplicates
- Add transaction wrapper to resource type creation
- Add comprehensive tests for defensive fix behavior


tgvashworth:
@claude Implement the review feedback here. Refactor the logic around `CampaignListResource` creation into something like `ensure_campaign_list_resources`.

PR #1272, opened 2026-01-06
feat: Log campaign actions from gang/list screen

### LINKED ISSUE #994: Log campaign actions from gang page
It would be useful for users to be able to log campaign actions directly from their own gang page, rather than having to go back to the overall campaign page.

Could add an option here:

Image

fiveeightsix:
Adds a "Log Action" link to the "Actions" widget on the gang/list screen. When clicked, navigates to the existing "Log action" form, with the "Related gang" field set. Ensures return URLs are set correctly.

Closes #994.


tgvashworth (approved):
Great stuff, thank you!

Issue #1826, opened 2026-05-27
Kill-transfer inflates stash cost cache; gear re-valuation on death is undefined


tgvashworth:
## Summary

When a fighter dies in campaign mode, the kill handler transfers their gear to the stash and bumps the stash's cached cost by the equipment's value **as costed on the dying fighter**. That value can differ from what the same gear costs on the stash fighter, and the caches are never reconciled — leaving `stash_current` permanently out of sync.

## Confirmed in production

List `00102c77-6013-4130-84c9-c3adbbd3264d`, stash fighter `8f71fcc0-94d2-4a3c-9a05-b85256071057`:

```
stash_fighter.cost_int()  (fresh recompute) = 25     # Second Best 15 + Vast Bulk 10
stash_fighter.rating_current (cached)        = 70
list.stash_current          (cached)         = 70
list.dirty                                   = False  # never self-heals
```

The stash is cached **45¢ too high**. The two assignments actually on the stash recompute to 15 + 10 = 25.

## How it happens

Fighter-kill handler (`gyrinx/core/handlers/fighter/kill.py`):

- `equipment_cost = sum(a.cost_int() for a in to_transfer)` (line ~122) — valued **on the dying fighter**. Vast Bulk cost ~55¢ on Bozur (a Hive Scum with equipment-list context); the same mutation costs only **10¢** on the stash fighter (no context, `cost_override=None`).
- `propagate_from_fighter(stash_fighter, Delta(delta=equipment_cost, ...))` (line ~173-174) bumps the stash fighter's cached `rating_current` by that delta.
- `create_action(..., stash_delta=equipment_cost, ...)` (line ~199-213) increments `list.stash_current` by the same delta.

Neither path recomputes the stash fighter from its actual assignments, and nothing sets `dirty=True`, so `facts_from_db()` never reconciles. Later removals decrement by the stash-context cost, widening the gap. The 45¢ here = Vast Bulk's dying-fighter value (~55) minus its stash value (10).

## The general question (for the team)

For this specific case, the answer is "Vast Bulk should never have moved" — handled by making Mutations `persistent` (sibling issue). But the underlying mismatch is general: **what value should equipment carry when it moves to the stash on death?** This can happen for weapons too (which won't be made persistent), e.g. a weapon whose cost differs between the original fighter's equipment list and the stash.

Options:

[...]

Issue #1824, opened 2026-05-27
Mark the Mutations equipment category as persistent (keep mutations on dead fighters)


tgvashworth:
## Summary

The **Mutations** equipment category should be marked `persistent` so mutations stay with a fighter when they die, rather than transferring to the stash.

`ContentEquipmentCategory.persistent` (`gyrinx/content/models/equipment.py:65`) already drives this behaviour: the fighter-kill handler keeps persistent-category gear on the (now dead) fighter and only transfers non-persistent gear to the stash (`gyrinx/core/handlers/fighter/kill.py:100-114`).

Mutations are conceptually tied to the fighter they grew on — they shouldn't end up in the gang stash to be re-equipped onto someone else.

## Why now

Found while investigating a real list (`00102c77-6013-4130-84c9-c3adbbd3264d`). Fighter "Bozur" died and his **Vast Bulk** mutation transferred into the stash. Confirmed in production: the stash fighter holds a `Vast Bulk` assignment in the `Mutations` category.

Marking the Mutations category `persistent = True` would have kept Vast Bulk on Bozur and avoided the transfer entirely.

## Change

Content-data change: set `persistent = True` on the `Mutations` `ContentEquipmentCategory` (and consider any other categories that are conceptually fighter-bound, e.g. archetype/gene-smithing categories — worth a quick review while here).

## Related

Surfaced alongside two other stash issues from the same investigation (display filter for house-additional gear; kill-transfer stash cost-cache drift).

tgvashworth:
Related issues from the same stash investigation (list `00102c77-6013-4130-84c9-c3adbbd3264d`): #1824 (Mutations persistent), #1825 (stash hides house-additional gear), #1826 (stash cost-cache drift on kill-transfer).

PR #1830, opened 2026-06-05
fix: show house-additional gear on the stash card

### LINKED ISSUE #1825: Stash card hides house-additional gear (counted in cost, not displayed)
## Summary

House-additional gear (equipment in a category that is `restricted_to` one or more houses — e.g. **Mutations**) is **counted in the stash cost but never displayed** on the stash card.

## Detail

`ListFighter.wargear()` has a dedicated `is_stash` branch (`gyrinx/core/models/list.py:3034-3041`):

```python
def wargear(self):
    # For stash fighters, show all non-weapon gear regardless of restrictions
    if self.is_stash:
        return [
            e
            for e in self.assignments_cached
            if not e.is_weapon_cached and not e.is_house_additional   # <-- still filters
        ]
```

The comment says "show all non-weapon gear regardless of restrictions", but the code still filters `and not e.is_house_additional` (where `is_house_additional` == `equipment.category.restricted_to.exists()`, `gyrinx/content/models/equipment.py:506-511`).

Consequence: a non-weapon, house-restricted assignment on the stash is excluded from **both** `weapons()` (it's not a weapon) and `wargear()` (it's house-additional) — so it renders nowhere. But `cost_int` sums **all** `assignments_cached`, so its value still shows in the stash cost badge. The item is invisible and can't be reassigned/sold/deleted by the owner.

## Reproduction

List `00102c77-6013-4130-84c9-c3adbbd3264d` (prod). Stash badge shows 70¢; only "Second Best (15¢)" renders. The stash also holds a `Vast Bulk` mutation (Mutations category, `restricted_to = ['Underhive Outcasts']`) that is hidden.

## Suggested direction (needs investigation)

Dropping `and not e.is_house_additional` from the `is_stash` branch would make the stash show all non-weapon gear (matching the comment). **However, the knock-on effects aren't obvious** — `is_house_additional` filtering interacts with the gear-grouping / house-additional rendering elsewhere (`gyrinx/core/models/list.py:3098, 3188, 3254, 3377-3379`). Needs a careful look before changing, including how house-additional gear groups are rendered on the stash card and whether anything double-counts or regresses for normal fighters.

Note: if the Mutations category is made `persistent` (sibling issue), mutations specifically won't reach the stash, but other house-restricted non-weapon gear could still hit this display gap.

tgvashworth:
Closes #1825.

## Summary

Two changes shipped together because they came out of the same investigation:

1. **Display fix.** `ListFighter.wargear()`'s `is_stash` branch was filtering `is_house_additional`, so house-restricted, non-weapon assignments on the stash (e.g. Mutations) rendered nowhere even though `cost_int` still counted them. The stash card has no separate `house_additional_gearline_display` section like normal fighter cards do, so the filter just hid value. Drop the `and not e.is_house_additional` clause from that branch.
2. **Data repair via an admin maintenance page.** Investigation showed 281 persistent-category equipment assignments are sitting on stash fighters across 146 lists in prod — legacy of kills that happened before the kill handler enforced persistence. About 75 of those can be reliably traced back to the dying Fighter via the ListAction ledger (assignment.created within ±1s of an UPDATE_FIGHTER "killed" action on the same list — essentially same-transaction matches from `handle_fighter_kill`). For those we can move the assignment FK back to the (still-dead) Fighter and reconcile the stash cache.

Why an admin page and not just a management command: Cloud Run doesn't expose a shell. The CLI command stays useful for local testing and is now a thin wrapper around the same service module the admin view uses.

## What's on the branch

- `gyrinx/core/models/list.py` — drop `is_house_additional` filter from stash `wargear()`.
- `gyrinx/core/maintenance/persistent_stash.py` — service module: `find_candidates(list_id=None)` returns Candidate decisions (move / no_match / ambiguous / null_fighter / alive / archived / wrong_list); `apply(list_id=None, triggered_by=…)` performs the moves and reconciles caches.
  - Kill-action lookup is batched into one IN-query, so the dry-run is not N+1.
  - For each affected list: marks involved fighters dirty, `lst.facts_from_db(update=True)` to reconcile (incidentally heals #1826's cache drift on touched lists), and appends an audit `ListAction` with the observed `stash_delta`.
  - Guards: tight 1s match window; ambiguous ties (<0.1s apart) are rejected; alive/archived/null-fighter/wrong-list all skip.
- `gyrinx/core/models/backfill.py` + migration — `Backfill` model recording each apply run (operation, triggered_by, list_id_scope, status, summary JSON, error).
- `gyrinx/maintenance/` — new app holding the admin UI only:
  - `MaintenanceAdminSite` extends `admin.site.__class__` (so it chains onto `AnalyticsAdminSite`); installed app order in `settings.py` places this app after `gyrinx.analytics` so the chain composes.
  - `/admin/maintenance/` — index of available data repairs + recent backfill runs.
  - `/admin/maintenance/persistent-stash/` — dry-run table with optional `?list_id=…` scoping; POST applies and redirects to detail.
  - `/admin/maintenance/backfill/<uuid>/` — per-run summary with per-list detail and audit-action IDs.
  - Gang names in all three views are `target=_blank` links to `/list/<uuid>` so you can inspect the affected list in a new tab.
  - All three views are superuser-gated (the standard `admin_view` only enforces `is_staff`; this mutates data so it needs tighter control).
- `gyrinx/core/management/commands/migrate_persistent_stash_items.py` — thin CLI wrapper around the service for local testing.

## What this does NOT do

- Doesn't touch the ~204 items where provenance can't be proven (pre-tracking kills, list_building items, etc.). They stay on the stash, unchanged.
- Doesn't mark any category persistent — that's a content-data change (#1824).
- Doesn't fix the general stash cost-cache drift (#1826), though touched lists are incidentally healed by `facts_from_db(update=True)`.

## Test plan

- Display fix: `test_stash_wargear_includes_house_additional` — confirmed failing on main and passing with the fix.
- Service / command: 10 tests covering match-and-move, audit ListAction creation, recompute against the stale stash-fighter cache, dry-run no-op, all skip paths (no_match / alive / archived / ambiguous / null_fighter), and a query-count regression test asserting `find_candidates` is not N+1.
- Admin views: 6 smoke tests covering anon/staff/superuser auth, dry-run rendering, full POST apply with Backfill row creation + audit ListAction, and list_id scoping isolation.
- Full suite: 2650 passed, 13 skipped.
- Manual: seeded a positive scenario locally (Bozur killed via `handle_fighter_kill`, Heirloom Spike transferred to stash, category flipped persistent post-kill) and verified end-to-end through the admin UI — dry-run shows 1 would-move, Apply redirects to the backfill detail page, stash recomputes 25 → 0, Heirloom Spike appears on Bozur's card, audit `ListAction` recorded with `stash_delta = -25`.


tgvashworth:
@coderabbitai Done in f5ca72b9 — refactored to use `make_content_fighter` / `make_list` / `make_list_fighter` and added a negative `weapons()` assertion.

tgvashworth:
Triage of the latest review round:

**@coderabbitai / @copilot — addressed in fdd918cb:**
- Kill-action match query now requires `applied=True` (Copilot — `_fetch_kills_by_list`).
- `apply()` revalidates each move inside the per-list lock and skips stale moves with `moved_off_stash` / `fighter_state_changed` / `disappeared` reasons (CodeRabbit's cross-request race). New regression test `test_apply_revalidates_inside_lock_and_skips_stale_moves` uses `monkeypatch` to inject a stale Candidate snapshot.
- Ambiguous test no longer relies on wall-clock proximity — both kill-action timestamps are pinned to the assignment's `created` via `QuerySet.update()`. No more CI-loaded flake (Copilot).

**Already addressed in earlier commits / stale:**
- Tooltip wording and tooltip docstring (Copilot, fighter_card_stash.html:81 and test_stash_fighter.py:270): the tooltip + `is_in_persistent_category` property were reverted in 0916e659. The current PR doesn't claim provenance from the persistent-category flag.
- Audit ListAction missing owner (Copilot): the service writes `owner=triggered_by or lst.owner` (`persistent_stash.py:226`).
- PR description / file header references (Copilot): description updated to cover the full scope; command/test headers correctly cite #1825 with #1824 / #1826 listed as out-of-scope.

**Verified resolved in earlier commits (CodeRabbit auto-marked):**
- Dry-run no longer takes row locks (`find_candidates` is read-only; transactions only inside `apply()`).

PR #1898, opened 2026-06-20
fix: freeze transferred gear value on death to stop stash cost drift (#1826)

### LINKED ISSUE #1826: Kill-transfer inflates stash cost cache; gear re-valuation on death is undefined

tgvashworth:
## Summary

- When a fighter dies in campaign mode, their non-persistent gear transfers to the stash. The kill handler valued that gear in the **dying fighter's** context (its equipment list, house overrides, expansions) and added that amount to the cached stash total — but the new stash assignment re-priced itself in the **stash's** context, which has no equipment list. The two numbers never matched, nothing marked the stash for recompute, so the cached stash value drifted from reality and stayed wrong.
- This pins each transferred item's value via `total_cost_override` at the price it carried on the dying fighter. The value that leaves the gang's rating now equals the value that lands in the stash, and the stash fighter's own recomputed cost matches the cached total — no drift.
- The pin stays with the item if it's later re-equipped out of the stash, so gear keeps its frozen price rather than re-pricing to whichever fighter picks it up.

## Why

A weapon can cost different amounts depending on who holds it (a fighter's equipment list can discount it). On death the gear moved to the stash but the books were updated with the old price while the stash counted the new one. Freezing the price on transfer keeps a death from silently changing what the gang's gear is worth, and keeps the cached stash total honest.

## Test plan

- [x] `test_kill_freezes_transferred_equipment_value` — a Lasgun priced 5¢ on the fighter's equipment list (15¢ at base) transfers to the stash frozen at 5¢; `list.stash_current` equals the stash fighter's recomputed `cost_int()` (no drift).
- [x] `test_frozen_stash_value_survives_reassignment_out` — re-equipping that frozen item onto a fighter whose list would charge 12¢ keeps it at 5¢; stash and rating totals stay balanced.
- [x] Existing `test_kill_fighter.py` suite green (no-discount gear still transfers at base price).
- [x] Adjacent suites green: admin cost-cache recompute, stash fighter, persistent-stash migration, equipment sell/removal/cost-override, fighter lifecycle, equipment edit, purchases, campaign attributes.
- [x] No schema change (`total_cost_override` already exists); missing-migration check passes.

Closes #1826

### DIFF
--- a/gyrinx/core/handlers/fighter/kill.py
+++ b/gyrinx/core/handlers/fighter/kill.py
@@ -53,7 +53,11 @@ def handle_fighter_kill(
 
     The fighter's cost goes from X to 0, reducing rating by X.
     Equipment transfers from fighter to stash don't change overall wealth,
-    but equipment is preserved in the stash for potential re-use.
+    but equipment is preserved in the stash for potential re-use. The
+    transferred gear's value is frozen at what it cost on the dying fighter
+    (pinned via ``total_cost_override``) so it doesn't re-price in the stash's
+    context — keeping the value that leaves the rating equal to the value that
+    lands in the stash, and keeping the stash cost cache in sync (issue #1826).
 
     Equipment in a category flagged ``persistent`` (see
     ``ContentEquipmentCategory.persistent``) is the exception: it stays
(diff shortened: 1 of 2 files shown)


tgvashworth:
Closing: the total_cost_override mechanism is superseded by a per-component cost-pinning design (see #1826). Freezing the whole assignment total conflates the user-facing "must cost X" override with acquisition-price freezing, and drifts when components are later purchased. The regression tests here will be rewritten against the new mechanism.

Issue #1855, opened 2026-06-09
refactor: maintainability debt from June 2026 audit (tracking issue)


tgvashworth:
Tracking issue for the maintainability findings from the June 2026 codebase audit (detailed reports in `.claude/notes/codebase-assessment-2026-06/`). Each finding now has its own issue:

**Structural**
- [ ] #1858 — split `core/models/list.py` into a domain package *(do first; unblocks the rest)*
- [ ] #1859 — bring `views/pack.py` under the handler layer, de-duplicate paired views
- [ ] #1863 — small mechanical cleanups (FactsMixin, cost-change signal factory, urls.py split, PackItemFormMixin)

**Unfinished migrations**
- [ ] #1860 — finish the facts/cost-cache migration, delete the legacy regimes
- [ ] #1861 — retire legacy statline + legacy advancement-application paths

**Design-level**
- [ ] #1862 — de-leak `VirtualListFighterEquipmentAssignment`
- [ ] #1864 — invert the pack-visibility default (design discussion)
- [ ] #1865 — single lifecycle status enum for `ListFighter`

**Frontend**
- [ ] #1866 — converge the four fighter-card template variants, hoist repeated template logic
- [ ] #1867 — inline-JS convention violations + three `index.js` bugs

Security fixes: PR #1857. Performance fixes: separate PR in progress. Test findings tracked in #1856.

PR #1904, opened 2026-06-20
feat: verify_facts_vs_live read-only drift harness (#1860 Stage 0)

### LINKED ISSUE #1860: refactor: finish the facts/cost-cache migration and delete the legacy regimes
From the June 2026 codebase audit (domain report — the single deepest source of accidental complexity).

Four cost/cache regimes currently run simultaneously:
1. live `cost_int()` compute,
2. per-request `cached_property` + signal invalidation,
3. the persisted facts/dirty tree (`rating_current` etc. on assignment→fighter→list),
4. ListAction deltas (`rating_before/after`, `applied`, `skip_apply`).

They're stitched together by `facts_with_fallback()` (explicitly labelled "MIGRATION PERIOD ONLY"), `can_use_facts` prefetch-sniffing, `check_wealth_sync()` drift telemetry ("temporary and will be removed"), and a deprecated `List.cost_int_cached` (#1215) that raises in DEBUG. Every cost-touching feature must reason about all four regimes; the known drift bugs (stash cost inflation, stale rating after kill) live in these seams.

**Proposal:** finish the migration and delete the old paths:
1. Use the existing drift telemetry to confirm facts vs live-compute agreement in prod.
2. Make ListAction a **pure audit log** — it never applies deltas; remove `skip_apply` plumbing.
3. Delete `facts_with_fallback`, `check_wealth_sync`, `cost_int_cached`, and the List-level cached properties (~600 lines).
4. The facts tree + `set_dirty` + `refresh_list_facts` task becomes the single cache layer with a single invalidation path (per the conventions rule: "Only ONE system should update cached values").

tgvashworth:
**Stage 0 of #1860 — a read-only diagnostic harness only. No behaviour change, no deletions.** Draft / parked: does NOT close #1860; the risky regime-collapse work (Stages A–D) is for a dedicated effort.

Adds `manage verify_facts_vs_live`: for every non-archived List, compares the **persisted** stored facts (`rating_current`/`stash_current`/`credits_current` — what the UI shows) against a **fresh live `cost_int()` recompute** (fresh instances, so per-request `@cached_property` can't mask drift). Reports each disagreeing list with both values + delta + dirty flag; exits non-zero on any mismatch. Writes nothing. Args: `--list-id`, `--owner`, `--include-archived`, `--quiet`.

This is the plan's Stage-0 gate: *prove facts == live before deleting any fallback.*

### Design correction worth noting
The plan suggested reading the persisted side via `facts_from_db(update=False)`. That **masks List-level drift** — at the List level it lazily re-derives the aggregate from child caches, so a stored List-field drift is silently re-derived away and never reported. This harness reads the **raw stored fields** instead (the faithful "what the user sees vs the truth" check) while recomputing the full assignment→fighter→list subtree on the live side, so drift at any level surfaces.

### Finding (run against the curated dev template `gyrinx_main`)
**5 of 72 lists already show drift** — including one List with `rating_current` persisted=0 vs live=235 (which the `facts_from_db` approach would have hidden). Concrete evidence the regimes disagree in practice. The dedicated #1860 effort should run this against prod as the real Stage-0 gate.

### Verification
- `pytest -n auto`: 2817 passed, 13 skipped (4 new tests)
- `manage check` clean; `makemigrations --check` no-op; `./scripts/fmt.sh` clean; `bandit` 0 issues

Part of #1855.

### DIFF
--- a/gyrinx/core/management/commands/verify_facts_vs_live.py
+++ b/gyrinx/core/management/commands/verify_facts_vs_live.py
@@ -1,0 +1,240 @@
+"""Read-only drift harness: compare each List's persisted facts to a live recompute.
+
+Stage 0 of #1860 (finish the facts/cost-cache migration). Before any of the
+legacy cost regimes can be deleted we must *prove* the persisted facts cache
+(``rating_current`` / ``stash_current`` / ``credits_current``) agrees with a
+clean live ``cost_int()`` recompute, list-wide. This command does exactly that
+and NOTHING else — it computes, compares, and reports. It never writes.
+
+For every non-archived ``List`` (optionally filtered by id or owner) it:
+
+  * reads the PERSISTED facts — the stored ``rating_current`` /
+    ``stash_current`` / ``credits_current`` fields. These are exactly what
+    ``List.facts()`` serves on a clean list and what the UI displays; they are
+    the values the migration intends to trust. (We read the raw stored fields,
+    not ``facts_from_db(update=False)``: the latter lazily *re-derives* the
+    List aggregate from the child fighter caches, so a List-level cache drift
+    would be silently re-derived away and never reported. The stored fields are
+    the thing we must prove correct.) and
+  * computes a fresh LIVE recompute by summing ``cost_int()`` over a brand-new
+    set of model instances (``@cached_property`` survives ``refresh_from_db``,
+    so we must fetch fresh instances to get a genuinely uncached recompute —
+    see the project memory note on this).
+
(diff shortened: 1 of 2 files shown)


tgvashworth:
Closing as superseded by #1976–#1979, which completed #1860. The production estate was reconciled during that work, and the balance-sheet debug view now provides the stronger ongoing diagnostic. Thanks—this Stage-0 harness and its drift findings helped inform the migration.

Issue #1863, opened 2026-06-09
refactor: small mechanical cleanups — FactsMixin, cost-change signal factory, urls.py split, PackItemFormMixin


tgvashworth:
From the June 2026 codebase audit (architecture report). Small mechanical refactors, low risk, each ~1-2 days:

- [ ] **`FactsMixin`**: the facts/`set_dirty`/`facts_from_db`/`can_use_facts`/`debug_facts_in_sync` pattern is triplicated verbatim across `List`, `ListFighter`, `ListFighterEquipmentAssignment` (15 method defs). Extract one mixin so the "only one system updates caches" invariant has one implementation.
- [ ] **`register_cost_change(sender, field)` factory** in `content/models/signal_handlers.py`: 10 near-identical ~18-line `pre_save` cost-change detectors + 10 matching `post_save` receivers, differing only in sender and field name (~200 lines collapsible).
- [ ] **Split `core/urls.py`** (1,011 lines, 204 routes, 70 commits/yr) via `include()` per domain (`urls/list.py`, `urls/fighter.py`, `urls/campaign.py`, `urls/pack.py`). Pure mechanical; mostly a merge-conflict win.
- [ ] **`PackItemFormMixin`** in `forms/pack.py` (1,742 lines, 19 classes): every `Content*PackForm` repeats the same Meta boilerplate and a near-identical ~12-line `clean_name` uniqueness check, plus repeated "group choices into Default/Custom by pack membership" `__init__` logic. Several hundred lines collapsible.

tgvashworth:
**Codebase-grounding pass (verified against current `main`).** All four are independent — recommend **four separate PRs**; the `urls.py` split is the lowest-risk standalone one to start with.

### 1. FactsMixin
The method group across the three classes:
- `facts()` — `List` **515**, `ListFighter` **2537**, `ListFighterEquipmentAssignment` **4530** — bodies are **verbatim identical**.
- `set_dirty()` — 601 / 2574 / 4542 — **nearly identical** (propagate up the chain).
- `facts_from_db()` — 617 / 2591 / 4561 — **differ** in body (List recurses into fighters/assignments; assignment defers to `cost_int()`). Don't naively merge these.
- `can_use_facts` / `debug_facts_in_sync` exist only on `List` (1236/583) and `ListFighter` (2550/2560).
**Start by extracting the verbatim-identical `facts()` + `set_dirty()`; leave `facts_from_db` as an overridable hook.**

[...]

PR #1909, opened 2026-06-20
refactor: extract register_cost_change signal factory (#1863 part 2)

tgvashworth:
Part 2 of #1863: extract a `register_cost_change(sender, field)` factory in `content/models/signal_handlers.py`, collapsing the 10 duplicated pre_save cost-change detectors + 10 post_save fan-out receivers into one factory + 10 registrations. **No behaviour change** (−159 lines).

- All 20 `dispatch_uid`s preserved verbatim (passed explicitly, not derived — legacy strings aren't one scheme); `@traced` span names unchanged; pre-change snapshot / enqueue / recompute untouched. `sync_auto_equipment_cost` (the genuinely-different ContentFighter post_save) left hand-written.
- **GC-safety fix:** Django connects receivers as weakrefs; factory closures (unlike the original module-level functions) had no global ref and were being garbage-collected (only the last registration survived). Kept alive via a module-level strong-ref list — preserves the original `weak=True` semantics. Verified across `gc.collect()`.
- `pytest -n auto`: 2813 passed; cost-signal suites (`test_content_cost_signals`, `test_propagate_content_cost_change`, `test_cost_propagation`, `test_list_refresh_cost`) 58 passed. `manage check` / `makemigrations --check` clean.

Part of #1855.

### DIFF
--- a/gyrinx/content/models/signal_handlers.py
+++ b/gyrinx/content/models/signal_handlers.py
@@ -38,221 +38,82 @@ logger = logging.getLogger(__name__)
 
 
 # =============================================================================
-# Pre-save signal handlers
+# Cost-change signal factory
 #
-# These signals detect when cost fields change on content models and mark
-# affected core objects (assignments, fighters, lists) as dirty.
+# Every content model that carries a cost field needs the same pair of signal
+# handlers: a pre_save detector that marks affected core objects dirty when the
+# cost changes, and a post_save receiver that enqueues the CONTENT_COST_CHANGE
+# fan-out. The handlers differ only in the sender model, the cost field name,
+# and the (legacy, non-uniform) dispatch_uid / @traced span names. We pass those
+# in explicitly rather than deriving them: the existing strings do not follow a
+# single scheme (e.g. ContentFighter's pre is "content_fighter_base_cost_change"
+# but its post is "content_fighter_cost_action"), and the dispatch_uid strings
+# must be reproduced verbatim — they dedupe signal connections, so any drift
+# risks a double-registered handler (double credit charges) or a disconnect-by-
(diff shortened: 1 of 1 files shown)


tgvashworth:
Closing as superseded — the *intent* is still valid, the *diff* isn't.

Since this branched (2026-06-20) main has moved 86 commits, and `content/models/signal_handlers.py` specifically gained **+419 / −116 lines across 9 PRs** — cost-cache deletion Stages B–D (#1977, #1978, #1979), unconditional recording (#1981), cost-pinning Phases 6–7 (#1941, #1944), promotions (#2019), fixture-load signal skip (#1995), durable task backend (#1985). Hence CONFLICTING. Rebasing a "collapse these 10 handlers into one factory" refactor across a rewrite of those same handlers is more work — and more risk — than redoing it.

The refactor is still worth doing: the duplication is all still there (10 `pre_save` detectors + 10 `post_save` fan-outs), and the handlers are now *more* uniform than when this was written — every post_save fan-out is identical, and the pre_save detectors gained the `_old_cost_for_propagation` snapshot uniformly.

Redoing it fresh against main. **Branch `issue-1863-signal-factory` is kept** — the two things worth carrying forward from it:

1. **The GC-safety fix.** Django connects receivers as weakrefs, so factory-created closures (unlike module-level functions) get garbage-collected and only the last registration survives. Needs a module-level strong-ref list to keep `weak=True` semantics. This fails *silently* with green-looking tests, so it's the trap to avoid on any re-attempt.
2. **The dispatch_uid list** — all 20 preserved verbatim (legacy strings aren't one scheme, so they must be passed explicitly, not derived).

Still tracked under #1863 / #1855.

Issue #1867, opened 2026-06-09
fix: inline-JS convention violations and three index.js bugs (URL-driven state)


tgvashworth:
From the June 2026 codebase audit (frontend report). The URL-driven UI state convention is well-followed in most of the app (deep-linkable filters, zero modals, dedicated confirmation pages) — but ~834 lines of inline JS across 27 templates contain the violations:

- **`campaign/property_schema_editor.html` + `sub_asset_schema_editor.html`** — 100% JS-dependent form builders writing JSON into `forms.HiddenInput`; the pages genuinely don't work without JS (worst offenders).
- **`pack_item_add.html`** — a checkbox clears/disables a select client-side: verbatim the forbidden pattern.
- **`print_config/form.html`** — radios disable *and uncheck* checkboxes client-side.
- **`list_fighter_advancement_type.html`** — JS rewrites cost inputs (server should render the right values per selection via navigation).
- **`index.js:445–475`** — global checkbox→hidden-field payload mutation; the handler never sets initial state despite its comment (pre-checked boxes double-submit).
- **`index.js:225–294`** duplicates server-rendered disable logic including a copy-pasted tooltip string.

Three additional plain bugs in `static/core/js/index.js`: malformed attribute selector at ~line 206 (only works via CSS auto-close at EOF), the initial-state bug above, and `form.submit()` at ~503 bypassing submit listeners/validation (should be `requestSubmit()`).

Also: Bootstrap JS loaded from CDN without SRI, with the version pinned in two places.

**Proposal:** fix the three JS bugs + SRI now; migrate the schema editors and the form-mutating templates to server-rendered URL-driven variants per the convention (each is a small, separable piece of work).

tgvashworth:
**Codebase-grounding pass (verified against current `main`).** Splitting this into the "do-now small" part vs the larger migrations.

### The "do-now" part — 3 JS bugs + SRI (one small PR)
`core/static/core/js/index.js`
1. **Malformed attribute selector — line 207**: `` `[data-gy-sync=${...}` `` is missing the closing `]` (only works because CSS tolerates the unclosed bracket at EOF). Fix the selector.
2. **Initial-state bug — lines 454–470**: the checkbox→hidden-field handler only attaches a `change` listener and never sets state on load, despite the `// Set initial state` comment. A pre-checked box double-submits until toggled. Set initial `disabled` on load.
[...]

Issue #1742, opened 2026-05-09
Archived content packs (and archived items) must keep showing up in subscribed lists/gangs


tgvashworth:
## Rule

Archiving a `CustomContentPack` (or a `CustomContentPackItem` inside it) is a soft-delete in the **pack owner's library**. It must NOT retract content from lists/gangs already subscribed to that pack.

If a list has a pack in its `packs` M2M, every item in that pack — including items whose `archived` flag is set, and packs whose `archived` flag is set — continues to be visible to that list. This applies to disciplines, powers, equipment, default assignments, skills, rules, and any other pack-aware content.

The same applies to `Campaign.packs`.

This rule has now been documented in `CLAUDE.md` under "Domain Rules → Content packs: archive semantics" so future work follows it.

## Background

- #1736 framed the inverse as a bug ("archived packs leak content into reads") and was closed as not-planned.
- #1735 added a filter on the psyker side that excludes archived items, which goes against this rule. That filter should be revisited.

## Audit / work items

Verify that pack-aware reads do NOT filter `archived=False` on `CustomContentPackItem` or `CustomContentPack`. In particular:

1. **`gyrinx/content/models/base.py` — `ContentQuerySet.with_packs()`**
   Currently filters `archived=False` on the inner `CustomContentPackItem` exists subquery. That filter should be removed so archived items in subscribed packs are still returned.

2. **PR #1735 psyker changes** — review the filter added there; it likely needs to be reverted or changed to match this rule.

3. **`gyrinx/core/views/list/views.py:289`** — builds `pack_content_map` from `CustomContentPackItem.objects.filter(pack__in=subscribed_packs, archived=False)`. The `archived=False` should be dropped so visual indicators include archived items.

4. **Search the codebase** for any other place that queries `CustomContentPackItem` with `archived=False` in a read-time context where a list/gang is being shown its subscribed content. Each one should be evaluated against the rule and removed if it filters out content the subscriber should still see.

5. **Pack management UI** — the rule does not change the pack owner's view of their own pack. Archived items can still be hidden in the pack admin/editor. Only the read path for subscribers is in scope.

## Test plan

- A list subscribed to a pack with an archived `CustomContentPackItem` for a custom skill/discipline/power/equipment must still see that item in its forms, fighter cards, and selection dropdowns.
- A list subscribed to an archived `CustomContentPack` must still see all of that pack's items.
- Cloning a list preserves its `packs` M2M (existing behaviour, unchanged).
[...]

PR #1910, opened 2026-06-20
refactor: extract PackItemFormMixin for Content*PackForm (#1863 part 3)

tgvashworth:
Part 3 of #1863: extract `PackItemFormMixin` in `core/forms/pack.py`, deduping the repeated `clean_name` uniqueness check + Default/Custom choice-grouping across 13 `Content*PackForm`s. **No behaviour change** (−38 net; the ~95-line mixin offsets removed boilerplate).

- 4 helpers (variants genuinely differ): `_raise_if_name_taken`, `_pack_item_object_ids`, `_check_name_unique_in_pack`, `_apply_grouped_pack_choices`. Each form keeps its own manager (`.filter` vs `.all_content()`), field scoping, and message text.
- **`archived=False` preserved** on every `CustomContentPackItem` lookup (matches the conditional unique constraint per the #1742 archive-semantics rule; documented in the mixin docstring with a CLAUDE.md citation).
- `pytest -n auto`: pack form/view suite 509 passed; full 2813 passed. `manage check` / `makemigrations --check` clean.

⚠️ Overlaps `forms/pack.py` with #1867 B1 (`issue-1867-b1`) — whichever merges second needs a trivial rebase.

Part of #1855.

### DIFF
--- a/gyrinx/core/forms/pack.py
+++ b/gyrinx/core/forms/pack.py
@@ -102,11 +102,16 @@ class PackItemFormMixin:
 
     CRITICAL — archive semantics (see CLAUDE.md "Content packs: archive
     semantics"): every ``CustomContentPackItem`` lookup here filters
-    ``archived=False``. This is a form-validation / unique-constraint write
-    path, and the ``CustomContentPackItem`` unique constraint is itself
-    conditional on ``archived=False`` — so the "live" item lookup MUST match.
-    Do not broaden these to include archived items.
+    ``archived=False``. Archiving is the pack owner's soft-delete, so an
+    archived item must stop reserving its name against its own owner —
+    otherwise they can never reuse a name they archived. Do not broaden these
+    to include archived items.
     """
+
+    # Forms that are never pack-scoped (and so never assign ``self._pack``)
+    # still inherit these helpers; the default keeps their "no pack" branches
+    # working rather than raising AttributeError.
+    _pack = None
 
     def _raise_if_name_taken(self, qs, message):
         """Exclude the current instance, then raise ``ValidationError`` if any
--- a/gyrinx/core/forms/pack.py
+++ b/gyrinx/core/forms/pack.py
@@ -81,6 +81,101 @@ def _append_cost_propagation_help(form, field_name="cost"):
         form.fields[field_name].help_text = existing + COST_PROPAGATION_HELP_SUFFIX
 
 
+class PackItemFormMixin:
+    """Shared helpers for the ``Content*PackForm`` family.
+
+    Collapses three repeated shapes without forcing a single ``clean_name``
+    (the variants genuinely differ — base vs ``all_content()`` manager,
+    extra-field scoping, per-pack uniqueness, and message text). Each form's
+    ``clean_name`` / ``__init__`` calls these helpers with its own specifics:
+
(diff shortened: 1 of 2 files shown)


tgvashworth:
Refreshed for current main (was 86 commits behind).

**Rebased cleanly** — no conflicts. `core/forms/pack.py` turned out to be byte-identical to this branch's base on main: the only two commits that touched it were #1867 B1 (#1911) and its revert (#1918), which cancel out. So the ⚠️ overlap warning in the description above is **moot** — B1 is no longer in main.

**Added tests** for the archive-semantics rule the mixin now centralises. `PackItemFormMixin` makes one place decide how archived pack items are treated across all thirteen `Content*PackForm`s, but nothing asserted it. New `test_pack_form_mixin.py` covers:

- `_pack_item_object_ids` excludes archived items (the shared primitive both the uniqueness check and the choice grouping build on);
- the user-visible consequence — an archived pack item must not reserve its name, since `CustomContentPackItem`'s unique constraint is itself conditional on `archived=False`. Without this the form rejects a name the database would accept, and a pack owner can never reuse a name they archived;
- the no-pack path is a no-op.

Verified by mutation: dropping `archived=False` from the mixin fails both archive tests.

Verification on current main: pack form/view suites 663 passed; full suite **3645 passed, 13 skipped**; `manage check` and `ruff` clean; no migrations.

tgvashworth:
Review round applied (`29c5ae9b`). An independent review pass checked all 13 forms line-by-line against the merge base and confirmed the no-behaviour-change claim holds — manager (`objects` vs `all_content()`), error strings, self-exclusion, extra scoping and which forms have an in-pack check are all preserved, and `archived=False` is neither broadened nor narrowed. Three things came out of it:

**Fixed — latent bug.** Six forms inherit the mixin without ever assigning `self._pack` (`ContentHouseForm`, `ContentRuleForm`, `ContentSkillCategoryPackForm`, `ContentGearPackForm`, `ContentWeaponPackForm`, `ContentPsykerDisciplinePackForm`). Inert today because they only call `_raise_if_name_taken`, but three helpers document a "no pack" branch that would instead raise `AttributeError`. Added a class-level `_pack = None`.

**Fixed — two extracted behaviours were unpinned.** Worth knowing: swapping Custom/Default in the grouping helper passed **all 736 pack tests**, and dropping the self-exclusion broke exactly one incidental test. Both are now single implementations shared by thirteen forms, so both are pinned, along with empty-group omission and the no-pack no-op.

**Fixed — docstring rationale was wrong.** `CustomContentPackItem`'s unique constraint is on `(pack, content_type, object_id)` — pack membership — not on the name, so my "rejects a name the database would accept" reasoning was misattributed. The actual justification is the soft-delete rule: an archived item must stop reserving its name against its own owner. Corrected in both the mixin docstring (flagged CRITICAL, so worth being right) and the test.

Verified by mutation — swapping grouping order, dropping self-exclusion, and removing the `_pack` default each fail the new tests. Full suite **3649 passed, 13 skipped**.

Acknowledged, not changed: the duplicated `self._pack is None` guard in `_check_name_unique_in_pack` leaves the `.none()` branch dead outside tests — it's defensive and covered, and removing a guard for tidiness isn't worth it in a no-behaviour-change PR.

PR #1911, opened 2026-06-20
refactor: make pack fighter statline-override URL-driven (#1867 B1)

tgvashworth:
#1867 B1: make the pack fighter **statline-override** server-rendered / URL-driven (works without JS). Removes the forbidden checkbox-disables-select inline JS.

- `?override_statline=1` navigation; the view passes the flag to `ContentFighterPackForm`, which includes `statline_type` only when overriding; the template renders an `<a>` toggle + a hidden input to carry the flag through the create POST. Fighter-only feature; the 2-step add flow's Step 2 is unaffected.
- New test: GET with/without `?override_statline=1` asserts `statline_type` present/absent in `form.fields` and rendered HTML; JS-off variants confirmed via test client.
- `pytest` pack suites 354 passed; `manage check` / `makemigrations` / `fmt` clean.

⚠️ Overlaps `forms/pack.py` with #1863 part 3 (`issue-1863-form-mixin`) — whichever merges second needs a trivial rebase.

Part of #1855 (#1867).

### DIFF
--- a/gyrinx/core/forms/pack.py
+++ b/gyrinx/core/forms/pack.py
@@ -153,26 +153,18 @@ class ContentFighterPackForm(forms.ModelForm):
 
     Accepts an optional ``pack`` kwarg to filter house/rules querysets
     to include both base library content and pack-specific content.
+
+    Whether the statline-override section is shown is a URL-driven variant:
+    the view reads ``?override_statline=1`` and passes ``override_statline``
+    to this form. When off, the ``statline_type`` field is removed entirely
+    so it can never submit a value (no client-side disabling needed).
     """
 
-    override_statline = forms.BooleanField(
-        required=False,
-        label="Override default statline for this Category?",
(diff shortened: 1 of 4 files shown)


tgvashworth:
## Manual browser test — PASS ✅

Tested on a dev server (Chrome via rodney, logged-in superuser). The statline override is genuinely URL-driven — the server renders the correct variant per URL, with no client-side form mutation.

**Override OFF (default, no query param)** — no "Statline type" select; the toggle is an outline link:

override off

**Override ON (`?override_statline=1`)** — server renders the statline-type select; toggle shows pressed; a hidden `override_statline=1` input carries the flag through the create POST:

override on

**The toggle is a navigation, not JS** — clicking it on the ON page drops the param and the select disappears:

toggle back

DOM-verified: OFF → select absent (not merely disabled), no inline toggle script remains; ON → select present + enabled. Works with JS off.

PR #1912, opened 2026-06-20
refactor: URL-drive advancement-type cost prefill (#1867 B2)

tgvashworth:
#1867 B2: URL-drive the advancement-type **cost prefill** (works without JS). Removes the `advancement_configs` JSON blob + the inline change-handler that rewrote cost inputs.

- `advancement_choice` is a GET-submitting select (`?advancement_choice=…`); the view sets the form's `initial` for `xp_cost`/`cost_increase` from the form's instance-level `advancement_configs` (so fighter-filtered stat + equipment advancements prefill correctly). Values stay editable. `<noscript>` button + auto-submit enhancement cover both paths.
- New tests (JS-off): GET `?advancement_choice=stat_strength` renders 8/30; no-param renders defaults; invalid choice falls back. Browser-smoke confirmed the JS path + cost round-trip.
- `pytest` advancement suites 81 passed; `manage check` / `makemigrations` / `fmt` clean.

Part of #1855 (#1867).

### DIFF
--- a/gyrinx/core/forms/advancement.py
+++ b/gyrinx/core/forms/advancement.py
@@ -235,7 +235,12 @@ class AdvancementTypeForm(forms.Form):
     advancement_choice = forms.ChoiceField(
         # Note that these choices are overridden by __init__()
         choices=ADVANCEMENT_CHOICES,
-        widget=forms.Select(attrs={"class": "form-select"}),
+        # data-gy-toggle-submit auto-submits the (GET) form on change so the
+        # server re-renders with the matching cost defaults; the page still
+        # works without JS via the explicit "Update costs" button.
+        widget=forms.Select(
+            attrs={"class": "form-select", "data-gy-toggle-submit": ""}
+        ),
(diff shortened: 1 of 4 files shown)


tgvashworth:
## Manual browser test — PASS ✅

Dev server, logged-in, campaign fighter with XP. Cost prefill is URL-driven — the server sets the per-choice `xp_cost`/`cost_increase` from the URL as form `initial`; no client-side input rewriting.

**Default load** — Willpower, XP 3 / Rating +5:

default

**No-JS proof (`?advancement_choice=stat_strength`)** — server renders Strength, XP 8 / Rating +30, purely from the URL:

strength

**JS on** — changing the select auto-navigates to `?advancement_choice=stat_attacks` (XP 12 / Rating +45); it's a real navigation, not an input rewrite:

attacks

Per-choice costs match `ADVANCEMENT_CONFIGS`; an invalid choice falls back to defaults; costs stay editable.

PR #1913, opened 2026-06-20
refactor: make print-config fighter selection mode URL-driven (#1867 B3)

tgvashworth:
#1867 B3: make the print-config **fighter-selection mode** server-rendered / URL-driven (works without JS). Removes the destructive inline JS that disabled + unchecked + dimmed the fighter checkboxes.

- `?fighter_selection_mode=…` navigation; the form is built for the resolved mode and includes `included_fighters` only for `specific`; `clean()` remains the source of truth (clears the M2M for non-specific modes, so switching specific→all on edit clears it even though the field is omitted). `<noscript>` submit + non-destructive auto-submit enhancement. Covers create + edit.
- 13 new view tests (create+edit, GET/POST across modes, incl. specific→all clearing the M2M and zero-fighter validation).
- `pytest` print-config 20 passed; full core suite 2376 passed; `manage check` / `makemigrations` / `fmt` clean.

Part of #1855 (#1867).

### DIFF
--- a/gyrinx/core/forms/print_config.py
+++ b/gyrinx/core/forms/print_config.py
@@ -52,9 +52,20 @@ class PrintConfigForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         self.list_obj = kwargs.pop("list_obj", None)
+        # The selection mode is URL-driven: the view passes the mode selected via
+        # navigation so the form renders the matching variant. The
+        # ``included_fighters`` field is only present when the mode is "specific".
+        self.selection_mode = kwargs.pop("selection_mode", PrintConfig.ALL_FIGHTERS)
         super().__init__(*args, **kwargs)
 
-        if self.list_obj:
+        # Keep the stored mode in sync with the URL-driven variant so the hidden
+        # field that posts the value matches the section being rendered.
+        self.fields["fighter_selection_mode"].initial = self.selection_mode
+
+        if self.selection_mode != PrintConfig.SPECIFIC_FIGHTERS:
+            # Only the "specific" variant shows (and submits) fighter checkboxes.
+            del self.fields["included_fighters"]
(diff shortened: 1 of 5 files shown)


tgvashworth:
## Manual browser test — ⚠️ BLOCKER (do not merge as-is)

The **headline behaviour works**: the server renders the fighter-checkbox section only for `specific` mode, per URL, with no destructive client-side mutation.

**`all` mode (default)** — no fighter section:

all mode

**`?fighter_selection_mode=specific`** — server renders the fighter checkboxes from the URL (the no-JS proof):

specific mode

**But the form can't actually be submitted** — saving returns 200 with four hidden "This field is required" errors and creates nothing:

submit fails

### Root cause (regression vs main)
The mode picker is a `<form method="get">` **nested inside** the main `<form method="post">`. Nested forms are invalid HTML, so the browser splits the form, which:
1. **orphans the "Create/Update Configuration" submit button** (it belongs to no form → clicking does nothing); and
2. parses the Blank Cards fields (`blank_fighter_cards`, `blank_vehicle_cards`) **outside** the main form → every submit fails validation with "This field is required" ×4 (the errors sit inside the collapsed Blank Cards accordion, so they're invisible).

So a PrintConfig **cannot be saved from this page in a real browser**. CI is green only because the new tests POST a complete dict via the Django test client (no HTML parsing) — false confidence.

This also violates the project convention (`gyrinx/core/templates/CLAUDE.md`): variant pickers must be `<a>` links carrying the new state in the query string (canonical: `house_rule_form.html`), not a nested `<form>`. Note B1 (#1911) does this correctly with `<a>` toggle links.

**Fix:** replace the nested GET picker with server-rendered `<a>` links (matching B1's pattern), then re-verify a real-browser save.

tgvashworth:
will come back to this

PR #1922, opened 2026-06-29
feat: shared campaign admins (arbitrators)

### LINKED ISSUE #988: Allow multiple admins on campaigns
As many campaigns operate with multiple admins, it would be great if campaign creators could set other users to have admin rights. I don’t think we need several levels of control — full shared admin should be fine.

broly5556-ship-it:
Campaigns previously had exactly one administrator: the owner. This PR lets the owner delegate full administrative rights to other users — "Arbitrators" — so several people can run a campaign together.

- Adds an `admins` many-to-many to `Campaign`, editable on the campaign settings form. Choices are restricted to campaign participants (gang owners) plus anyone already granted admin; the owner is always an admin and cannot be removed.
- Everywhere that previously authorised "campaign owner only" now asks `campaign.is_admin(user)`: the campaign management views (edit, add/remove gangs, start/end/reopen/archive, assets, resources, attributes, packs, copy-in, captured fighters, battles), the fighter and gang arbitrator screens (injuries, state, XP, credits, cost refresh) via the shared `Permission.ARBITRATOR` seam in `views/fighter/permissions.py`, and the templates via a single `is_admin` context flag.
- The campaign header lists the arbitrators, and roster changes are recorded in the campaign action log ("Arbitrators updated: added …; removed …").
- `is_admin()` is memoised per instance and prefetch-aware, so per-fighter-card template checks don't add queries.

Built on @broly5556-ship-it's original implementation of the model, migration, and form layer.

Closes #988

## Test plan

- [x] Unit and view tests in `gyrinx/core/tests/test_campaign_admins.py` — `is_admin` semantics, admin-gated campaign views exercised as a shared admin and as a stranger, arbitrator permission on fighter views, battle creation, form choice restriction, roster audit log, and template controls
- [x] Full suite passes locally
- [x] Manual browser test: granted an arbitrator via the settings form, confirmed owner-parity controls, started the campaign as the shared admin (budget distribution and action-log attribution correct), and used the injuries screen on another player's gang
- [ ] CI green (fork workflows need maintainer approval)

## How to try it

1. As a campaign owner, open the campaign → Edit and tick a participant under **Arbitrators**.
2. Sign in as that user: the campaign page shows Start/End, Edit, Add Gangs, asset/resource/attribute/pack management, and the per-gang row actions — all functional, not just visible.
3. Open another player's gang in the campaign: the arbitrator fighter controls (injuries, XP, state) work there too.

**Behaviour note:** shared admins get full parity, including editing the arbitrator roster (per #988's "full shared admin"). The owner can never be removed.


tgvashworth (commented):
Thanks so much for picking this up, @broly5556-ship-it — #988 has been on our wish list for a while and it's great to see a PR for it. The model, migration and form wiring all look clean, and I appreciate you working through the earlier bot feedback too.

There's one thing that needs sorting before this can land, plus a few smaller points I've left inline.

**The main one: the feature isn't wired through to the views yet.** The template changes correctly reveal the admin-only controls (Add Gangs, Remove from Campaign, Start/Edit campaign, asset/resource/attribute management, copy-in) to shared admins — but none of the backing views were updated to *authorise* those actions. Every one of them still gates on `get_object_or_404(Campaign, id=id, owner=request.user)` (or an equivalent owner-only check), and the new `Campaign.is_admin()` method isn't actually called anywhere outside its unit test. So as it stands a shared admin sees all the new buttons, but clicking any of them gives a 404 or a "you don't have permission" redirect — the feature looks enabled but can't actually be used.

To close the gap, the views that currently do the owner-only lookup need to consult `campaign.is_admin(request.user)` instead. There's a spread of them across `views/campaign/lists.py`, `lifecycle.py`, `assets.py`, `resources.py`, `attributes.py`, `copy.py` and `crud.py`, plus the inline check in `campaign_remove_list`. A view/integration test that acts *as an admin user* against one or two of those endpoints would also catch this whole class of regression — the current unit test only exercises `is_admin()` in isolation, which is why CI is green despite the actions not working.

The smaller points (the `admins` form field exposing every user on the site, the permission expression being duplicated ~9× in the templates instead of a single context flag, the un-prefetched `admins.all` queries, and the test's location/fixtures) are all inline below and are much lighter lifts.

Really nice start though — the hard modelling decisions are already made here, it just needs the authorisation half to catch up with the UI half. Happy to help if any of the view-layer plumbing is unclear.

gyrinx/core/models/campaign.py
```diff
@@ -154,12 +162,11 @@ def can_reopen_campaign(self):
         return self.status == self.POST_CAMPAIGN
 
     def is_admin(self, user):
```
tgvashworth, line 186: This is the key piece. `is_admin()` is defined and unit-tested, but it's never actually called by any view. The permission checks in the campaign views — e.g. `edit_campaign`, `campaign_add_lists`, `campaign_remove_list`, the lifecycle start/end/reopen/archive views, and the asset/resource/attribute-type CRUD views — all still do `get_object_or_404(Campaign, id=id, owner=request.user)` or check `request.user != campaign.owner`. Because of that, a shared admin sees the controls the templates now expose but gets a 404 / permission-denied when they use them.

The fix is to have those views authorise via `campaign.is_admin(request.user)` instead of the owner-only lookup. Grepping for `owner=request.user` under `gyrinx/core/views/campaign/` should surface them all.

gyrinx/core/templates/core/campaign/campaign.html
```diff
@@ -233,7 +242,7 @@ <h2 class="h5 mb-0">Gangs</h2>
                         <a href="{% url 'core:campaign-attributes' campaign.id %}"
                            class="link-primary link-underline-opacity-50 link-underline-opacity-100-hover fs-7">Manage Attributes →</a>
                     {% endif %}
-                    {% if campaign.owner == user and not campaign.is_post_campaign and not campaign.archived %}
+                    {% if (campaign.owner == user or user in campaign.admins.all) and not campaign.is_post_campaign and not campaign.archived %}
```
tgvashworth: Concrete example of the gap: this now shows "Add Gangs" to shared admins, but `campaign_add_lists` in `gyrinx/core/views/campaign/lists.py` does `get_object_or_404(Campaign, id=id, owner=request.user)`, so an admin who isn't the owner gets a 404 when they click it.

gyrinx/core/templates/core/campaign/includes/campaign_lists.html
```diff
@@ -64,7 +64,7 @@
                                     {% endif %}
                                 {% endfor %}
                                 <td class="text-end pe-0">
-                                    {% if campaign.owner == user or list.owner == user %}
+                                    {% if (campaign.owner == user or user in campaign.admins.all) or list.owner == user %}
```
tgvashworth: Same pattern here — this exposes "Remove from Campaign" to shared admins, but `campaign_remove_list` guards with `if request.user != campaign.owner and request.user != list_to_remove.owner:` and redirects with a permission error otherwise. An admin who isn't the list owner hits that error, so this check needs an admin branch too.

gyrinx/core/forms/campaign.py
```diff
@@ -135,6 +134,7 @@ class Meta:
             "summary",
             "narrative",
             "public",
+            "admins",
```
tgvashworth, line 145: `admins` is added to `fields` but there's no widget or queryset restriction, so it falls back to Django's default `ModelMultipleChoiceField(queryset=User.objects.all())` with a bare `<select multiple>` — that renders every user account on the site (unstyled, and enumerable by any campaign owner). Worth constraining the queryset (e.g. to campaign participants) and giving it a `form-select`-styled widget to match the rest of the form.

gyrinx/core/templates/core/campaign/campaign.html
```diff
@@ -60,7 +60,7 @@ <h1 class="h2 mb-0">{{ campaign.name }}</h1>
                                 {{ star_count|default:0 }}
                             </button>
                         </form>
-                        {% if campaign.owner == user %}
+                        {% if campaign.owner == user or user in campaign.admins.all %}
```
tgvashworth: This `campaign.owner == user or user in campaign.admins.all` expression is repeated ~9 times across this template and `campaign_lists.html`. We already compute owner-style booleans in the view and pass them as context (`context["is_owner"]` in `views/campaign/views.py`, and the same pattern in `assets.py` / `attributes.py` / `resources.py` / `packs.py`). Could you compute a single `is_admin` flag in the view — it can just call the new `campaign.is_admin(user)` method — and read that here instead? Keeps the permission logic in one place and avoids the parenthesisation drifting between the copies. CLAUDE.md has this as a convention: "Compute derived/conditional display logic in the view as a flag; templates just read it."

gyrinx/core/templates/core/campaign/includes/campaign_lists.html
```diff
@@ -119,7 +119,7 @@
                                 </td>
                             {% endfor %}
                             <td class="text-end pe-0">
-                                {% if campaign.owner == user or list.owner == user %}
+                                {% if (campaign.owner == user or user in campaign.admins.all) or list.owner == user %}
```
tgvashworth: Because `admins` isn't prefetched on the campaign queryset, each `campaign.admins.all` in the templates is a fresh, uncached query — and this one sits inside the per-gang `{% for %}` loop, so it re-queries once per row. A `prefetch_related("admins")` on the campaign fetch in the detail view (and/or the single-context-flag approach from the other comment) collapses it back to one query.

gyrinx/tests/test_campaign_admins.py
```diff
@@ -0,0 +1,32 @@
+import pytest
```
tgvashworth: Two small things on the test. It lives in `gyrinx/tests/`, but the other campaign tests are under `gyrinx/core/tests/` (`test_campaign.py`, `test_campaign_actions.py`, etc.) — worth moving it there for consistency. And our conftest already has `make_user` and `make_campaign` fixtures; CLAUDE.md asks us to use those rather than calling `create_user` / `Campaign.objects.create` inline. A follow-up test that drives one of the admin-gated *views* as an admin user would also be really valuable — that's the bit that would have caught the authorisation gap.

tgvashworth:
Following up on my review with the bigger-picture steer — this one changes the *shape* of the fix rather than just adding to it, so I wanted to flag it clearly.

The key thing to know: the app already has an "arbitrator" concept, and right now it's defined as *the campaign owner*. It lives in [`gyrinx/core/views/fighter/permissions.py`](https://github.com/gyrinx-app/gyrinx/blob/48d2c9ab25306c07437995bd8e1dd749cd0f9413/gyrinx/core/views/fighter/permissions.py#L19-L26) — there's a `Permission.ARBITRATOR`, plus a single helper (`get_user_permissions` / `get_list_and_fighter`) that every arbitrator-gated screen already calls to ask "can this person run this campaign?" That one function is what governs fighter kills/injuries, XP awards, pack management, and the campaign settings.

Your `Campaign.is_admin()` is really asking the same question — "who's allowed to arbitrate this campaign?" So rather than it living as a separate notion that only the templates consult, the direction we'd like to take this is to **fold the new admins list into that existing definition**. Concretely:

- teach `get_user_permissions` ([permissions.py:24](https://github.com/gyrinx-app/gyrinx/blob/48d2c9ab25306c07437995bd8e1dd749cd0f9413/gyrinx/core/views/fighter/permissions.py#L19-L26)) that a user is an arbitrator if they're the owner **or** in `campaign.admins`, and
- have the campaign views authorise through `campaign.is_admin(request.user)` instead of their current owner-only `get_object_or_404(..., owner=request.user)` lookups.

The payoff is that it does the whole job in one place: the moment shared admins count as arbitrators, every screen you'd want them to have — kill/injure fighters, award XP, manage packs, edit settings, add/remove gangs — starts working for them automatically, instead of us wiring up (and testing) each view by hand. It also collapses the duplicated `owner == user or user in campaign.admins.all` checks in the templates down to a single flag.

Completely understand why `permissions.py` didn't jump out — it's tucked under `views/fighter/`, which isn't the obvious home for "who runs a campaign." But that's the seam this should plug into. The modelling you've already done (the M2M + migration + form) is a solid base for exactly this — it's really just a matter of connecting it to the permission check that already exists rather than running alongside it.

Happy to point out the specific call sites that need to move over to the helper if that'd make it easier to get started.

tgvashworth:
Thanks again for the groundwork here, @broly5556-ship-it — I've pushed a follow-up commit to your branch (maintainer edits) to close the gap from my review and get this over the line, since the modelling you did was the right base to build on.

What the new commit does:

- **Wires `is_admin()` through the views.** Every campaign view that gated on `owner=request.user` (add/remove gangs, lifecycle, assets, resources, attributes, packs, copy-in, edit) now authorises via a shared `get_campaign_admin_or_404()` helper, so the controls the templates reveal actually work for shared admins.
- **Folds shared admins into the existing arbitrator concept** in `views/fighter/permissions.py` — so admins can also run the arbitrator screens (fighter injuries/state, XP, credits, battles) without per-view wiring.
- **Single `is_admin` context flag** replaces the repeated `campaign.owner == user or user in campaign.admins.all` template expressions, and `is_admin()` is memoised/prefetch-aware so the per-fighter-card paths don't pick up N+1 queries.
- **Constrains the form field**: arbitrators are picked from campaign participants (plus existing admins) with a checkbox widget, rather than a bare select over every user on the site.
- **Migration regenerated** at the head of the graph (0146 → 0167) after merging main in to resolve the conflicts.
- **Tests moved** to `gyrinx/core/tests/`, switched to the conftest fixtures, and extended with view-level coverage acting as a shared admin and as a non-admin.

Full suite passes locally. Once CI is green I'll get this merged. Thanks for kicking it off — this has been a long-requested feature (#988).

gyrinx/core/models/campaign.py
tgvashworth: Done in 2c342c5b + a1ac9826. Every campaign view that did the owner-only lookup now authorises via `get_campaign_admin_or_404()` (new helper in `views/campaign/common.py`) or an inline `campaign.is_admin()` check — lists, lifecycle, assets, resources, attributes, packs, copy-in, crud, sub-assets, captured fighters, action logging, and battle creation. Shared admins also count as arbitrators in `views/fighter/permissions.py`, so the fighter injuries/state/XP/credits screens work for them too. View tests in `gyrinx/core/tests/test_campaign_admins.py` drive the endpoints as a shared admin and as a stranger.

gyrinx/core/templates/core/campaign/campaign.html
tgvashworth: Fixed in 2c342c5b — `campaign_add_lists` now fetches via the admin-aware helper, so the Add Gangs flow works for shared admins. Covered by `test_add_gangs_view_allows_shared_admin` (200 as admin, 404 as stranger); also verified in the browser.

gyrinx/core/templates/core/campaign/includes/campaign_lists.html
tgvashworth: Fixed in 2c342c5b — `campaign_remove_list` now checks `campaign.is_admin(request.user)` alongside the list-owner branch. Covered by `test_remove_list_allows_shared_admin`.

gyrinx/core/forms/campaign.py
tgvashworth: Fixed across 2c342c5b / af6778b4 / a1ac9826. The queryset is now campaign participants (gang owners) plus already-granted admins, excluding the owner — no site-wide user enumeration. It renders with the house `BsCheckboxSelectMultiple`, and an unbound form gets `User.objects.none()` so the all-users fallback can't come back. `test_edit_form_admins_choices_limited_to_participants` covers it.

gyrinx/core/templates/core/campaign/campaign.html
tgvashworth: Fixed in 2c342c5b — the views now compute a single `is_admin` flag (`campaign.is_admin(user)`) and the templates just read it; the owner-or-admin expression is gone from the templates entirely. For the include-heavy list-side gates (fighter cards, list header) there's an `is_campaign_admin` template filter backed by the memoised `is_admin()`.

gyrinx/core/templates/core/campaign/includes/campaign_lists.html
tgvashworth: Fixed in 2c342c5b — the per-row checks read the context flag now, the detail view prefetches `admins` for the header list, and `is_admin()` memoises the admin ids per instance (and uses the prefetch cache when present), so repeated calls cost at most one query per campaign instance.

gyrinx/tests/test_campaign_admins.py
tgvashworth: Done in 2c342c5b + a1ac9826 — moved to `gyrinx/core/tests/test_campaign_admins.py`, rewritten on the conftest fixtures (`make_user`, `make_campaign`, `make_list`, `list_with_campaign`), and expanded to 12 tests including the view-level coverage you asked for: edit/add-gangs/remove-list/injuries/battle-new driven as a shared admin and as a stranger, plus the form restriction, roster audit log, and page controls.

gyrinx/core/templates/core/campaign/campaign.html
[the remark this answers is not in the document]
tgvashworth: Fixed in 2c342c5b — the archived-state unarchive branch and the other listed blocks all read the `is_admin` context flag now; no owner-only template gates remain in campaign.html / campaign_lists.html.

tgvashworth:
Closing the loop on the bigger-picture steer above: implemented exactly as described in 2c342c5b — `get_user_permissions` / `get_list_for_edit` in `views/fighter/permissions.py` now treat a user as `Permission.ARBITRATOR` when they're the campaign owner **or** in `campaign.admins` (via a reusable `arbitrator_q()`), and the campaign views authorise through `campaign.is_admin(request.user)` instead of the owner-only lookups. That gave shared admins the fighter kill/injury/XP/credits screens and battle management with no per-view wiring, as predicted.

PR #1931, opened 2026-07-03
test: run the balance-sheet matrix across a catalog-vs-pack content axis (#1826)

### LINKED ISSUE #1930: Pack-scoped content price changes never mark assignments dirty or create audit actions
## Summary

Editing the cost of any **pack-scoped** content item (equipment, weapon profile, accessory, upgrade — anything registered via `CustomContentPackItem`) silently does nothing to the lists using it: no assignments are marked dirty, no `CONTENT_COST_CHANGE` action is created, no campaign credit adjustment happens. Gangs keep displaying the old price until some unrelated recompute snaps their caches — with no audit trail.

## Mechanism

`get_old_cost()` (`gyrinx/content/signals.py:28`) reads the pre-save price via `model_class.objects.only(cost_field).get(pk=...)` — the **default ContentManager**, which excludes pack content. For a pack row this raises `DoesNotExist`, the helper returns `None`, and every cost-change signal handler in `gyrinx/content/models/signal_handlers.py` interprets that as "new instance, no existing assignments" and returns without calling `set_dirty()` or flagging `_cost_changed` — so the async propagation task is never enqueued either.

Notably the *downstream* task was already pack-hardened (`gyrinx/core/tasks.py` re-fetches via `all_content()` with an explicit comment) — the upstream signal just never fires for pack rows.

## Reproduction

Pinned as the strict-xfail **P7** group in the balance-sheet situation matrix (`gyrinx/core/tests/test_balance_sheet.py`, pack params of the two content-price-change cells): change a subscribed pack item's cost, and the harness shows nothing dirtied, then cache-vs-recompute divergence with no action.

## Fix direction

Mirror the task's pattern in `get_old_cost()`: use `all_content()` where the manager provides it, falling back to the default manager. One line plus flipping the P7 xfails. Scheduled with the Phase 3 producer fixes in the #1826 programme.

Found by the pack-vs-catalog axis added to the situation matrix.

Refs #1826

tgvashworth:
## Summary

Content that lives in a content pack is fetched differently from catalog content throughout the app — the default manager hides it, and only `all_content()`/`with_packs()` see it. That means every cost path has a pack-shaped way to go wrong that tests using only catalog content can never catch (the previous PR's only blocking review finding was exactly such a bug).

- **Every balance-sheet matrix cell now runs twice** — once with catalog content, once with identical content inside a subscribed content pack — via a parametrized fixture and a small gear factory, not duplicated tests. The calibration meta-tests double too.
- **First run found a real bug (#1930) — fixed here:** changing the price of a pack item did nothing — no gang marked for recalculation, no history entry, no campaign credit adjustment. The pre-save signal read the old price through the pack-excluding manager, concluded the row was new, and stopped. Fixed with the same `all_content()` fallback the async task already uses; both expected-failure tests flipped to passing on both sides.
- **The dead accessories-edit form branch is deleted here too.** It rewrote a weapon's accessories with no cost bookkeeping at all (catalog-side drift), and couldn't remove pack accessories anyway (silent no-op). No template posted to it; a bare POST now just re-renders the page, with a both-sides test pinning the inert behaviour.
- Cells whose *expected* behaviour genuinely differs by side use explicit per-side parameters with per-side expected-failure marks, keeping the strict tripwire discipline intact.
- **The independent review then caught the axis green-stamping a bug** (#1933): pack upgrades are booked at **zero** — the upgrade-stack sum excludes pack rows from its own total, and the recalculation path is equally blind, so the books "agree" at the wrong number and pure reconciliation can't see it. Healthy cells now also assert the **actual booked movement** (rating changed by the component's price); that turned the vacuous green into two strict expected-failure tests (buy-upgrade pack side, and reassignment of gear carrying a pack weapon profile). The fix touches the hot prefetch path and ships with Phase 3.

## Test plan

- [x] Full matrix on both sides: 59 passed, 8 expected failures (P2 kill-transfer, P3 sale, P4 stash removal on both sides; P8 pack-upgrade zero-booking ×2; P5 and P7 fixed and flipped).
- [x] Full suite: 2885 passed, 13 skipped, 8 xfailed. Accessory/signal/pack suites: 313 passed.
- [x] Design doc updated: seventh producer, the axis written into §5.2, counts corrected.

Refs #1826
Refs #1933
Closes #1930

### DIFF
--- a/.claude/notes/cost-pinning-design.md
+++ b/.claude/notes/cost-pinning-design.md
@@ -54,9 +54,9 @@ stash visibly contains.
 **Cache-vs-recompute divergence (the drift class).** Any mechanism that makes
 `cost_int()` disagree with the delta-accumulated cache is invisible until the next
 recompute snaps the cache to a different number — no error, just wealth jumping.
-Seven producers of this class existed when the harness landed; two (the
-pack-sweep gap and the bare-form endpoint) were fixed in the pack-axis PR,
-leaving five live:
+Eight producers of this class have been found so far; two (the pack-sweep
+gap and the bare-form endpoint) were fixed in the pack-axis PR, leaving six
+live:
 
 - **#1925 — component purchase under `total_cost_override`.** Buy an accessory on an
   assignment that has the override set. The handler computes the accessory's live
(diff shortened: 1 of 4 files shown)


tgvashworth:
Re CodeRabbit's nitpick: deliberate — that cell's accessory comes from its own pack, separate from the axis pack, to prove cross-pack behaviour; keeping the explicit registration.

PR #1937, opened 2026-07-03
feat: in-place component through models + inert pin columns (#1826 Phase 4)

tgvashworth:
## Summary

Phase 4 of the cost-pinning programme: **the storage for price memory, changing no behaviour.** Once later phases fill these columns, gear will remember where its price came from — which fixes the last outstanding bookkeeping bug (death-transfer repricing, the original #1826) without breaking price-correction propagation.

- **The three component relationships (weapon profiles, accessories, upgrades) become real models — on their existing tables.** A state-only migration (`SeparateDatabaseAndState`, zero database operations) declares through models matching the join tables Django already created: same names, same BigAuto keys, same columns handler code already queries, same unique constraints. No parallel tables, no data copying, no cutover later. Every existing `.add()`/`.set()`/`.remove()`/`.all()` call site works unchanged.
- **A second, purely additive migration adds the pin columns, all empty**: per component row and for the assignment's base cost — a link to the price-setting content row, the amount at acquisition, and an explicit state (`unpinned`/`source`/`catalog`/`derived`/`orphaned`; everything existing is `unpinned`). The state enum exists because a bare null link would be ambiguous between three different meanings — the design review caught that ambiguity as a future silent-repricing bug.
- **The assignment admin moves to per-row inlines** (Django hard-refuses plain multi-select widgets over explicit through models — startup check `admin.E013`); pin columns show read-only per row, giving staff pin visibility for free.
- **Proof of neutrality**: the entire books-checker matrix is untouched and green; the hot-path query snapshot count is unchanged at 64, with the only textual diffs being the four new assignment columns in SELECT lists (verified mechanically, not by eyeball); `manage check` and migration checks clean.

## Test plan

- [x] New `test_pin_schema.py`: add/set/remove/clear through the declared models; rows born `unpinned` with null amounts; the handler-level through-table filter pattern still works; duplicate rows still coalesce; admin change form + changelist smoke tests.
- [x] Both migrations applied against a real database forked from production data (state-only first — no DDL — then additive columns); `makemigrations --check` clean afterwards.
- [x] Full suite: 2912 passed, 13 skipped, 2 xfailed (the deliberate kill-transfer tripwires).

## Next

Phase 5: resolution honours a pinned amount when present (still a no-op — every amount is null until acquisition writes begin in Phase 7).

Refs #1826


tgvashworth:
CodeRabbit's nitpick applied in fcf3493a — base-cost pin fields now visible read-only on the assignment admin, matching the through-row inlines.

PR #1944, opened 2026-07-04
feat: acquisitions write pins; clone carries them (#1826 Phase 7)

tgvashworth:
Part of the cost-pinning programme (#1826), Phase 7 of the design in `.claude/notes/cost-pinning-design.md` (§4.6). Phases 1–6 are merged: resolution reads pinned amounts, and sweeps maintain them when content is corrected. Until now, nothing ever *wrote* a pin — all that machinery has been idle. This phase turns it on.

## What changed

**One choke point writes the acquisition receipt.** `pin_assignment()` (new, `gyrinx/core/cost/pinning.py`) runs the same lookups live resolution uses, in the same precedence order, and records on each row: the amount paid, the price-setting source as an FK (expansion item or equipment-list row), and the pin state — SOURCE for override-priced gear, CATALOG for book price, DERIVED for computed amounts (formula accessories, cumulative upgrade stacks). It is idempotent: only unpinned rows are written and an existing receipt is never overwritten, so it is safe to call from any layer, and a call on a legacy assignment is a value-neutral early instalment of the Phase 8 backfill.

**Every live acquisition path calls it**: the equipment purchase handler (covering the buy flow — the view creates the assignment, the handler pins before its first cost read), the accessory/profile/upgrade purchase handlers, `ListFighter.assign()`, equipment advancements, and vehicle purchases. Deliberately *not* wired: default-kit materialisation and linked-child creation (structurally free — the resolution anchors outrank any pin), the admin write surface (permanent policy exception, remediated by the recompute action), and the death transfer (Phase 9 converts it to the clone-with-pins path; the P2 expected-failure cells keep tracking it).

**Cloning carries receipts.** `clone()` now copies the base pin fields and creates component through-rows with their pin fields intact — a cloned gang (including the campaign-start clone) keeps its prices instead of silently reverting to live pricing. The two expected-failure tripwires from Phase 6 flip to real passing assertions. Copying via the through rows also fixes a latent hole where cloning dropped pack-scoped profiles and upgrades (the old loops iterated pack-filtering managers).

**A CI guard makes missed paths loud.** A watchdog test scans non-test source for the three assignment-creation shapes and compares against a labelled inventory (pinned / copies / anchored / kill-until-Phase-9 / admin / content-side). A new acquisition path that forgets to pin fails CI with instructions; verified to trip on a synthetic bypass site.

## Deliberate behaviour changes

Two harness cells asserting pre-pin behaviour split into pinned + legacy variants:

- **Mid-window recomputes are now pin-stable** for newly-bought gear: between a content edit and the async task, a recompute reads the receipt and the books hold their value; the correction lands only with the task, which rewrites the amount and books the delta. (Legacy unpinned rows keep the old visible-gap behaviour until the Phase 8 backfill.)
- **Reassignment is price-neutral for pinned gear** — the programme's signature outcome, arriving for all newly-acquired gear: a discounted lasgun moved to a fighter who would price it at catalog keeps its discounted price, no reprice telemetry fires, and the books reconcile. (Legacy rows still reprice, still covered.)

Death-transfer gear still reprices (kill path unwired until Phase 9), and existing gear stays receipt-less until Phase 8.

## How to verify

- `pytest gyrinx/core/tests/test_pinning.py` — per-source provenance (catalog / equipment-list / expansion / legacy-preference / formula / cumulative), anchors staying unpinned, value-neutrality (`cost_int()` identical before and after pinning), no-overwrite, and one receipt test per wired path.
- `pytest gyrinx/core/tests/test_pin_guard.py` — the watchdog.
- `pytest gyrinx/core/tests/test_balance_sheet.py` — the matrix, including the flipped clone cells and the new pinned/legacy variant pairs.
- Full suite: 2990 passed, 2 xfailed (the P2 death-transfer pair, Phase 9).

Refs #1826.

### DIFF
--- a/gyrinx/content/models/signal_handlers.py
+++ b/gyrinx/content/models/signal_handlers.py
@@ -60,6 +60,9 @@ def handle_equipment_cost_change(sender, instance, **kwargs):
     new_cost = get_new_cost(instance, "cost")
     if old_cost != new_cost:
         instance._cost_changed = True  # Flag for post_save to create actions
+        # The pre-change value rides to the async task: amount-snapshot
+        # (DERIVED) receipts are maintained by delta, which needs it.
+        instance._old_cost_for_propagation = old_cost
         instance.set_dirty()
 
 
@@ -78,6 +81,9 @@ def handle_fighter_base_cost_change(sender, instance, **kwargs):
     new_cost = get_new_cost(instance, "base_cost")
     if old_cost != new_cost:
         instance._cost_changed = True  # Flag for post_save to create actions
+        # The pre-change value rides to the async task: amount-snapshot
+        # (DERIVED) receipts are maintained by delta, which needs it.
+        instance._old_cost_for_propagation = old_cost
         instance.set_dirty()
 
 
@@ -96,6 +102,9 @@ def handle_profile_cost_change(sender, instance, **kwargs):
     new_cost = get_new_cost(instance, "cost")
     if old_cost != new_cost:
         instance._cost_changed = True  # Flag for post_save to create actions
+        # The pre-change value rides to the async task: amount-snapshot
+        # (DERIVED) receipts are maintained by delta, which needs it.
+        instance._old_cost_for_propagation = old_cost
         instance.set_dirty()
 
 
@@ -127,6 +136,9 @@ def handle_accessory_cost_change(sender, instance, **kwargs):
 
     if changed:
         instance._cost_changed = True  # Flag for post_save to create actions
+        # The pre-change value rides to the async task: amount-snapshot
+        # (DERIVED) receipts are maintained by delta, which needs it.
+        instance._old_cost_for_propagation = old_cost
         instance.set_dirty()
 
 
@@ -145,6 +157,9 @@ def handle_upgrade_cost_change(sender, instance, **kwargs):
     new_cost = get_new_cost(instance, "cost")
     if old_cost != new_cost:
         instance._cost_changed = True  # Flag for post_save to create actions
+        # The pre-change value rides to the async task: amount-snapshot
+        # (DERIVED) receipts are maintained by delta, which needs it.
+        instance._old_cost_for_propagation = old_cost
         instance.set_dirty()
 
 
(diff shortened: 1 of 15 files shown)


tgvashworth:
Deep-review pass complete. Outcome, fixed in 8457964f + c6edc01e:

**HIGH — the Phase 6 sweeps contradicted the cumulative-stack receipts Phase 7 writes** (reviewer-reproduced). Phase 7 pins SINGLE-mode upgrade stacks at the override-inclusive cumulative walk; Phase 6's sweep re-derived those receipts to the *raw catalog* sum — a +5 rung correction rewrote a 26¢ receipt to 35¢ and booked +9¢, silently destroying the acquisition discount on a rung that was never corrected. And corrections to the per-rung override itself never reached the receipt (amount-snapshots carry no FK). **Corrections now apply by delta**: the receipt moves by exactly what the corrected contribution moved. The pre-change value rides from the pre_save handler through the task payload; holders whose per-rung override masks a catalog change keep their receipt untouched; direct callers without the pre-change value leave receipts standing and flag the snapshot fallback rather than guess. Moved rows keep the documented v1 amount-snapshot imprecision — per-rung provenance is the escalation if it bites. Four regression cells built on real receipts.

**MEDIUM — balance sheet never reported `pricing == "pinned"`** — owner-spotted an hour before the review landed; fixed in 8457964f (base line classifies pinned/override/live with the pin state as detail; aggregate lines classify pinned/mixed/live by their rows).

**LOW — pinning under a fixed total** wrote a "receipt" at prices that were never paid and would have blocked the Phase 8 freeze-conversion (receipts are never overwritten). Pinning now skips frozen assignments entirely.

**Guard hardened**: creation patterns cover `get_or_create`/`update_or_create`/`bulk_create`, reverse-manager creates, and direct through-model creates; PINNED-labelled files must contain their expected `pin_assignment` call counts (deleting a call now fails CI); the synthetic-bypass negative case is a real test; heuristic limits documented.

**Test gaps closed**: profile-level expansion pins, and the zero-anchor paths driven through the real `convert_default_assignment` flow.

**Confirmed clean by the review**: provenance fidelity resolver-by-resolver (precedence, legacy tie-break, expression base input, from-default membership, cost coercion), value-neutrality, wiring order and rollback safety (insufficient-credits rolls pins back atomically), clone through_defaults (no m2m_changed receivers exist to miss; the pack-manager sidestep holds), and every pin state lining up with a sweep domain.

Suite: 3000 passed, 2 xfailed (P2 death-transfer, Phase 9).

Issue #1947, opened 2026-07-04
Background tasks: registration is only enforced in prod — make it impossible to forget


tgvashworth:
## Summary

Declaring a background task is a two-step ritual split across two files with nothing coupling them: `@task` in `gyrinx/core/tasks.py`, plus a `TaskRoute` entry in `gyrinx/tasks/registry.py`. Only the **production** Pub/Sub backend enforces the second step (enqueue lookup + topic provisioning) — dev's ImmediateBackend never consults the registry and tests call `.func` directly. Miss step two and everything is green locally while the task is dead on arrival in prod.

This just happened: `backfill_pins` (PR #1946, the Phase 8 cost-pinning backfill) shipped unregistered and was caught in review, not by any tooling. Every future task author walks the same trap.

**Full analysis:** https://gist.github.com/tgvashworth/d954a346a7f96905a8f142714eb1a73e

## Prioritized recommendations (all three)

1. **Django system check (do first).** Import the known task modules, collect every `django.tasks.Task` instance, and **error** if any is missing from the registry — and in the other direction, error on registry entries whose function no longer exists. Runs on `runserver`/`migrate`/`check --deploy`/CI, so the mistake turns into a red error in the author's terminal seconds after it's made. ~40 lines, zero prod behaviour change.

2. **Declaration-owned registration (the abstraction).** A thin `@gyrinx_task(ack_deadline=..., schedule=...)` decorator that applies Django's `@task` **and** registers the `TaskRoute` as a side effect of declaration. `registry.py` shrinks to a list of task *modules* to lazily import (keeping the existing circular-import-safe shape); per-task entries disappear. Forgetting becomes structurally impossible for tasks in known modules; the check from (1) polices the rare new-module case. Migration is mechanical — move each `TaskRoute` kwarg onto the decorator; `get_task`/`get_all_tasks` keep their signatures so `backend.py` and `provisioning.py` don't change.

3. **Dev backend parity.** Wrap ImmediateBackend so `enqueue()` consults the registry before delegating — dev then fails exactly like prod at the first real enqueue. Doesn't cover `.func`-style test calls, which is why it complements rather than replaces (1).

## Rejected

Full auto-discovery (no registry at all): the registry deliberately carries per-task config (ack deadlines, cron schedules) and is the explicit provisioning surface for Pub/Sub topics — infrastructure with cost and IAM surface shouldn't be created as an implicit side effect of an import.

## Key code points

[...]

PR #1970, opened 2026-07-11
feat: system check that every @task is registered

tgvashworth:
## Why

Declaring a background task is two steps in two files with nothing coupling them:

1. `@task` in the task module (e.g. `gyrinx/core/tasks.py`), and
2. a `TaskRoute` entry in `gyrinx/tasks/registry.py`.

Only the **production** Pub/Sub backend enforces step 2 (it looks the task up to find its topic, and provisions topics from the registry). Dev's `ImmediateBackend` never consults the registry, and tests call `.func` directly — so a task that's declared but not registered is **green everywhere locally and dead on arrival in prod**. This exactly happened with `backfill_pins` (the Phase 8 cost-pinning backfill), caught in review rather than by any tooling.

## What this adds

A Django system check (`gyrinx/tasks/checks.py`) that runs wherever checks run — `runserver`, `migrate`, `check --deploy`, CI — and:

- **E001** — errors if a function declared with `@task` in a known task module is **not** in the registry. This is the actual mistake, turned into a red error in the author's terminal seconds after they make it.
- **E002** — errors if a registry entry isn't a real `@task`-decorated function.

It scans the configured `TASK_MODULES` **plus** any module that already owns a registered task, so "added a second task beside a registered one" is caught even if the module list wasn't touched. **Zero production behaviour change.**

This is **recommendation 1 of the three** in the issue (the one marked "do first"). Recommendation 2 (a `@gyrinx_task` decorator that registers on declaration) and 3 (dev-backend parity) remain as separate follow-ups.

## Test plan

- [x] `manage check` runs the check and passes against the current registry.
- [x] New tests (`gyrinx/tasks/tests/test_checks.py`): current registry is clean; a dropped registration surfaces E001 for the right task; a non-`@task` registry entry surfaces E002; the check is actually registered with Django; `_declared_tasks` attributes tasks to their defining module.
- [x] `pytest gyrinx/tasks/tests/` — 107 passed.

Refs #1947

### DIFF
--- a/gyrinx/tasks/apps.py
+++ b/gyrinx/tasks/apps.py
@@ -30,6 +30,10 @@ class TasksConfig(AppConfig):
         # Import signal handlers to register them (works with any backend)
         from gyrinx.tasks import signals  # noqa: F401
 
+        # Import the system check module so its @register() runs — enforces that
+        # every @task is in the registry, in every environment (#1947).
+        from gyrinx.tasks import checks  # noqa: F401
+
         # Only provision Pub/Sub in Cloud Run environment
         if not os.getenv("K_SERVICE"):
             logger.debug("Not in Cloud Run, skipping task provisioning")
--- a/gyrinx/tasks/checks.py
+++ b/gyrinx/tasks/checks.py
@@ -1,0 +1,92 @@
+"""System check: every ``@task`` in a known module must be registered (#1947).
+
+Declaring a background task is a two-step ritual split across two files with
+nothing coupling them: ``@task`` in the task module, plus a ``TaskRoute`` entry
+in ``gyrinx/tasks/registry.py``. Only the **production** Pub/Sub backend enforces
+the second step (enqueue lookup + topic provisioning) — dev's ImmediateBackend
+never consults the registry and tests call ``.func`` directly. Miss step two and
+everything is green locally while the task is dead on arrival in prod.
+
+This check turns that omission into a red error at check time — ``runserver``,
+``migrate``, ``check --deploy``, CI — seconds after the mistake is made, in the
+author's own terminal. It also flags registry entries that aren't actually
+``@task``-decorated functions.
+"""
+
+import importlib
+
+from django.core.checks import Error, register
+from django.tasks import Task
+
+from gyrinx.tasks.registry import get_all_tasks
+
+# Modules that declare ``@task`` functions. Forgetting to register a task added
+# to one of these is the common trap this check closes. We also scan every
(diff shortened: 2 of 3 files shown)


tgvashworth:
Review round applied (`cb289572`).

**Implemented:**
- **Auto-discover `<app>.tasks` for every first-party `gyrinx.*` app** and fold it into the scan set — closes the main documented gap (a brand-new app's first, unregistered task is now caught, not just tasks added to already-known modules). Scoped to `gyrinx.*` so a third-party app's task module can't spuriously demand entries in our registry. `manage check` still clean (only `gyrinx.core.tasks` exists today, all registered).
- Use the public `TaskRoute.path` instead of the private `_underlying_func`.
- Skip an unimportable module path gracefully instead of crashing the check.
- Tests for the discovery path, the route-module derivation, and the graceful skip.

**Declined, with reasons:**
- **Duck-typing tasks via `.func`/`.enqueue` instead of `isinstance(obj, Task)`** (Copilot ×3): `_declared_tasks` scans *every* module global (`vars(module)`), so a loose `hasattr(obj, "func")` would false-positive on any global that happens to have a `.func` (a `functools.partial`, a property object, …). Sweeping a namespace needs a *precise* type check; `django.tasks.Task` is a public, exported class, so this isn't fragile coupling. (The `TaskRoute` duck-typing works because it operates on a *known* task, not a swept namespace.)
- **Keying by full path vs bare name** (duplicate task names): the registry itself keys by bare name (`get_task(name)`), so making the check stricter than the registry it validates would be inconsistent. That's a registry-level concern, out of scope here.

Also verified empirically that `route`'s module resolves to `gyrinx.core.tasks` (not `gyrinx.tasks.route`), so the derivation was correct — the new test pins it.

tgvashworth:
Both applied in `99781180`: emptied `TASK_MODULES` (auto-discovery covers `gyrinx.core.tasks` — `manage check` stays clean), and `_declared_tasks` now skips only a genuinely-absent target module, letting a broken module's import error surface instead of hiding it.

tgvashworth:
`07242394`: E001 hint now names the declaring module (`from <module> import <name>`) so the fix is copy/pasteable.

Re the E002 duck-typing suggestion — declining, as in the earlier round: `_declared_tasks` sweeps every module global, so it needs a precise `isinstance(obj, Task)` (a `.func` duck-check false-positives on partials/properties), and using the same precise check on both sides keeps one detection mechanism rather than two. `django.tasks.Task` is a public, exported class, so this isn't fragile coupling.

Issue #1949, opened 2026-07-05
Balance sheet: highlight fighter computed/cached mismatches; pin-source tooltip on pricing column


tgvashworth:
Two polish items for the balance-sheet debug view (`/_debug/list/<id>/balance-sheet/`), noticed while test-driving the Phase 8b maintenance flow:

1. **Fighter-level computed vs cached mismatches aren't highlighted.** When a fighter's computed total disagrees with their cached value, the row renders plainly — the mismatch is only visible by reading the two numbers. It should get the same red/danger treatment other problems get, so drift jumps out at a glance. (The `FighterBalance.is_mismatch` predicate already exists; this is template wiring.)

2. **Pricing column tooltip should describe the pin source.** The pricing cell currently shows `pinned (source)` / `pinned` / `mixed` / `live`. For pinned lines, a tooltip (or title attr) with a human-readable description of the attribution would make receipts self-explanatory — e.g. "Pinned to equipment-list price: Abyssal Ferrymen Ganger — Sweep Lasgun (5¢)" or "Catalog price at acquisition (15¢)". The design doc's §5.1 `source_repr` field was speced for exactly this and never implemented; the pin FKs are on the rows, so it's a resolver + template change.

Both are display-only — no behaviour or bookkeeping changes.

Refs #1826.

tgvashworth:
The two display-only polish items in the description above (mismatch-row highlight; pin-source tooltip on the pricing column) are implemented in #1975 (in review).

Keeping this issue **open as the umbrella for the broader pin-editing UI** — design §2.5 / "Phase 10": letting a user re-price a pin (audited), which is what makes a wrong/unwanted backfill or acquisition pin user-fixable. That's a separate, larger piece #1975 does not touch — the tooltip just makes pin attribution *visible*; pin *editing* is still to come.

PR #1975, opened 2026-07-11
feat: balance-sheet mismatch highlight fix + pin-source tooltip

tgvashworth:
Two display-only polish items on the balance-sheet debug view (`/_debug/list/<id>/balance-sheet/`), from #1949. No behaviour or bookkeeping changes.

## 1. Fighter/summary mismatch rows now actually highlight

When a fighter's computed total disagreed with its cached value the row rendered plainly. The `mismatch` class *was* applied — but `.fighter-row`'s blue background overrode `.mismatch`'s red (equal specificity, `.fighter-row` defined later in the stylesheet), so the highlight never showed. Bumped the rule to `tr.mismatch` (element+class) so the danger red wins on fighter and summary rows.

## 2. Pin-source tooltip on the pricing column

The pricing cell showed `pinned (source)` / `pinned` / `mixed` / `live` but nothing about *what* a pinned line is pinned to. Implemented the design doc's §5.1 `source_repr` (speced, never built) on `ComponentLine`:

- **SOURCE** → resolves the attribution FK: *"Pinned to Abyssal Ferrymen Ganger — Sweep Lasgun (5¢)"*
- **CATALOG** → *"Catalog price at acquisition (15¢)"*
- **DERIVED** / **ORPHANED** → describe their state
- user overrides name themselves
- Aggregate lines (profiles/accessories/upgrades) resolve a lone pinned row fully, else summarise the mix by state (*"Pinned: 2× catalog, 1× live"*)

Surfaced as a `title` tooltip on the pricing cell with a dotted-underline help cue.

## Test plan

- [x] New tests: `_source_repr` per pin state; `_rows_source_repr` single-row-resolve, mixed-summary, all-unpinned; `build_balance_sheet` populates a purchased line's `source_repr`.
- [x] `pytest gyrinx/core/tests/test_balance_sheet.py gyrinx/core/tests/test_debug_views.py` — 103 passed.
- [x] No model change → no migration.

Refs #1949

### DIFF
--- a/gyrinx/core/cost/balance_sheet.py
+++ b/gyrinx/core/cost/balance_sheet.py
@@ -84,7 +84,9 @@ def _source_repr(state, amount, fk=None) -> str:
     if amount is None or state == PinState.UNPINNED:
         return ""
     if state == PinState.SOURCE:
-        where = str(fk) if fk is not None else "an override price"
+        # A SOURCE pin should always name its FK; the fallback is defensive, and
+        # stays neutral so it can't be misread as a user cost-override.
+        where = str(fk) if fk is not None else "a price source"
         return f"Pinned to {where} ({amount}¢)"
     if state == PinState.CATALOG:
         return f"Catalog price at acquisition ({amount}¢)"
--- a/gyrinx/core/cost/balance_sheet.py
+++ b/gyrinx/core/cost/balance_sheet.py
@@ -106,7 +106,7 @@ def _rows_source_repr(rows) -> str:
     ]
     if not priced:
         return ""
-    if len(priced) == 1 and len(rows) == 1:
+    if len(rows) == 1:  # the sole row is priced (unpriced-only returned above)
         r = priced[0]
         return _source_repr(r.pin_state, r.pinned_amount, _pin_fk(r))
     counts = Counter(r.pin_state for r in priced)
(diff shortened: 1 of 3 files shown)


tgvashworth:
Review round applied (`4c7b3e10`):

- **Test made deterministic** — `test_balance_sheet_populates_pinned_line_source_repr` now asserts `base.pricing == "pinned"` before checking `source_repr`, so a pin-on-acquisition regression fails loudly rather than passing via the empty-else branch.
- Covered the SOURCE label in the multi-row summary branch; simplified a redundant clause in `_rows_source_repr`; corrected the CSS comment (`.summary` has no background rule, so only `.fighter-row` was overriding the red).

On the **N+1** (Copilot): acknowledged and left as-is. It only fires for SOURCE-pinned lines (CATALOG/DERIVED/ORPHANED have null FKs → no query), and this is a low-traffic staff/debug view. The cheap-looking fix — `select_related`-ing the four pin FKs — lives on the shared `with_related_data()` / `lst.fighters()` prefetch, so it would hit the main fighter-list view and force a `performance_view_queries.json` snapshot change: disproportionate blast radius for a debug tooltip. If the debug view ever gets slow on a large gang, that's the targeted fix.

Issue #1853, opened 2026-06-07
Tools of the trade (equipment sets) support


louisashworth:
We should introduce support for **equipment sets** — the "Tools of the Trade" rules. The core idea: a fighter with the *Tools of the Trade* special rule can maintain several different "cards", each fielding a different subset of the equipment and weapons they own. The fighter still owns everything on their sheet; each card just chooses which of those items are carried for a given battle.

This links with #1357 (crew selection) because every card created as part of a set is eligible for selection in a crew.

> **This ticket consolidates #1034** ("Option to hide weapons when a fighter has more than three", reported by @chrisshoto). That request — hiding weapons from the statblock when a fighter carries more than three, without dumping them to the stash — is a specific use case of equipment sets. It is folded in below as a primary use case, plus an optional weapon-carry-limit awareness enhancement. #1034 is closed as a duplicate of this ticket.

## Rules background (Core Rulebook pg. 85)

The *Tools of the Trade* rule lets a fighter have multiple cards based on different equipment sets. Key points that matter for us (highlighted paragraphs from the rulebook):

<img width="436" height="1038" alt="Tools of the Trade rules" src="https://github.com/user-attachments/assets/8fbad69e-c089-465e-8637-75de66bb2613" />

- It is **only equipment and weapons** that change between cards — we don't need to worry about most other things that can be attached (skills, rules, injuries, XP, etc. are shared across all cards).
- A weapon can **leave behind its attachments or special ammo** on a given card — so inclusion needs to be toggleable at the level of individual weapon accessories/upgrades/profiles, not just the weapon as a whole.
- Related carry rule (see #1034): a fighter may carry a **maximum of three weapons**, and **two-handed / Unwieldy weapons count as two** toward that limit. Equipment sets are how a player picks *which* three they field.

## The rating problem

Fighters show their overall cost on the list, and a gang's rating is the sum of fighter cost + their equipment. Once a fighter can field a subset of what they own, the rating of the *currently selected* sets can be lower than the *maximum possible* rating. That could confuse people, so the topmatter should show both — an overall (max) rating and, in brackets, the rating of the sets currently selected:

<img width="483" height="142" alt="Topmatter rating mock" src="https://github.com/user-attachments/assets/0dde9bbd-d4b5-458f-8aac-ac7635c9107f" />

And on fighter cards (we may want a better treatment here):

[...]

PR #1979, opened 2026-07-11
refactor: delete the legacy cost-sync scaffolding (#1860 Stage D)

### LINKED ISSUE #1860: refactor: finish the facts/cost-cache migration and delete the legacy regimes

tgvashworth:
Final stage of the cost/cache regime cleanup. Stages A–C (#1976, #1977, #1978) left the persisted cache fields with a single writer and one read path; this PR deletes the transitional scaffolding that guarded and cross-checked the old regimes while they coexisted.

## What changed

- **`FEATURE_LIST_ACTION_CREATE_INITIAL` removed.** The flag has shipped enabled since introduction and every production list has a bootstrap action (verified: 0 of 9,693 lists without one). The bootstrap CREATE action is now written unconditionally for new lists and clones, and the propagation guard reduces to "does the list have a `latest_action`".
- **`check_wealth_sync` deleted**, with its `list_cost_out_of_sync` telemetry and tests. It only fired from `List.cost_int()`, which detail pages no longer call; the balance-sheet debug view (`/_debug/list/<id>/balance-sheet/`) is the real instrument and checks strictly more (computed-vs-cached, credits ledger, action-chain continuity).
- **`debug_facts_in_sync` deleted** (List and ListFighter) along with the `fighter_debug.html` template and its three includes. The staff debug row in the list header now links to the balance sheet instead of showing the old red/green flag.
- **`List.rating` and `List.stash_fighter_cost_int` deleted.** These live-recompute cached properties were the last cost readers outside `cost_int()`/facts; nothing in templates read them. The #1853 selected-rating fallbacks now read `rating_current` directly.
- **`ListAction.applied` reads dropped** (reprice dedup exclude, kill-audit filter, balance-sheet field). The column stays and is now always written `True` — including the bootstrap CREATE, which previously fell through to the field default of `False`. Safety check against production: all 3,086 `applied=False` rows are bootstrap CREATEs; no row of any filtered action type ever had `applied=False`, so removing the filters cannot change what they match.
- **Dead `with_latest_actions()` prefetches removed** from the listing views (home, user profile, public lists, campaign add-lists picker). Those pages render `list_row.html`, which reads only cached fields — the prefetch ran a `DISTINCT ON` query per page for nothing. Detail pages keep the prefetch via `with_related_data`, which the performance snapshot covers (unchanged).
- **Tests:** ~90 tests de-parametrized off the flag (keeping the enabled-path assertions, which are now the only behaviour); the three dedicated flag-off tests deleted or repurposed — the fighter-edit one survives as a no-bootstrap-action test since `create_action` still returns `None` for lists outside the action system.

No migrations: everything deleted was Python-level (properties, methods, a settings flag); the `applied` column is untouched.

## Test plan

- [x] Full suite green: 3,158 passed, 13 skipped
- [x] Production data checks via read-only prodshell: `applied=False` rows are exclusively bootstrap CREATEs; 0 lists without a ListAction
- [x] Performance snapshot test unchanged and passing (detail-page prefetch retained)

Closes #1860


gyrinx/core/tests/test_kill_fighter.py
```diff
@@ -474,29 +472,24 @@ def test_kill_fighter_creates_list_action_with_stash_delta(
         .first()
     )
 
-    if feature_flag_enabled:
-        assert list_action is not None, "ListAction should be created when flag enabled"
+    assert list_action is not None, "ListAction should be created when flag enabled"
```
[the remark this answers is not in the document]
tgvashworth: Fixed in ec180b59.

gyrinx/core/tests/test_clone.py
```diff
@@ -528,28 +495,25 @@ def test_list_clone_actions_have_correct_deltas(
...
+
+    # The CREATE action records the CLONE's recalculated values, not the
+    # source's stale caches. The empty clone recalculates rating/stash to
+    # 0; credits are copied (500). Booking the source's 1000/150 here would
+    # contradict the clone's real cost and break the chain at the seam.
+    if result.cloned_action:
+        assert result.cloned_list.rating_current == 0
+        assert result.cloned_action.rating_delta == 0
+        assert result.cloned_action.stash_delta == 0
+        assert result.cloned_action.credits_delta == 500
```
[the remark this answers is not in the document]
tgvashworth: Fixed in ec180b59 — hard assert, block dedented.

gyrinx/core/tests/test_handlers_equipment_purchases.py
```diff
@@ -927,30 +858,27 @@ def test_handle_equipment_reassignment_stash_to_regular(
...
+    if lst.latest_action:
+        assignment.refresh_from_db()
+        stash.refresh_from_db()
+        fighter.refresh_from_db()
+        # Assignment rating_current should stay the same
+        assert assignment.rating_current == initial_assignment_rating
+        # Stash should lose the equipment cost
+        assert stash.rating_current == initial_stash_rating - equipment_cost
+        # Fighter should gain the equipment cost
+        assert fighter.rating_current == initial_fighter_rating + equipment_cost
```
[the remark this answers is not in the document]
tgvashworth: Fixed in ec180b59 — all 11 latest_action soft guards removed (this file + test_handlers_fighter_removal.py).

docs/technical-design/cost-propagation-architecture.md
```diff
@@ -496,11 +496,10 @@ def transact(
...
         )
     """
-    # Feature flag check (existing)
-    if not self.latest_action or not settings.FEATURE_LIST_ACTION_CREATE_INITIAL:
-        # Still execute mutation even if actions disabled
+    # Lists outside the action system (no bootstrap action) still mutate
+    if not self.latest_action:
         mutation()
-        # TODO: This should also preform the credit update as per current create_action
+        # TODO: This should also perform the credit update as per current create_action
```
tgvashworth, line 509: Do we need to worry about this?

gyrinx/core/cost/propagation.py
```diff
@@ -32,9 +31,9 @@ def has_change(self) -> bool:
 
 def _should_propagate(lst: "List") -> bool:
```
tgvashworth, line 32: Do any lists NOT have a `latest_action`? If not... this could be deleted, and much more code besides

docs/technical-design/cost-propagation-architecture.md
tgvashworth: No — this TODO is inside the design-era sketch of `transact()`, which was never built (there's no `transact` in the codebase; `create_action` kept its name and is now a pure record). In the shipped code, credits move via `spend_credits()` / `apply_credit_delta()`, which are **not** gated on `latest_action`, so there's no path where a mutation lands but its credit movement is skipped. Annotated the sketch as not-implemented in 1afcdcdc so it stops reading like a live gap.

gyrinx/core/cost/propagation.py
tgvashworth: In prod: no — 0 of 9,693 lists lack a ListAction (checked via prodshell during the Stage-0 gate), and after this PR both creation paths (creation + clone) write the bootstrap unconditionally, so the app can't mint new action-less lists.

What can still produce them: anything that bypasses `handle_list_creation` — raw ORM creation (tests do this deliberately via `make_list(create_initial_action=False)` and direct `List(...)`; shell seeding; any future script). The invariant is handler-enforced, not model-enforced.

If we accept 'a list's chain starts at its current caches on first mutation' (create_action already derives its before-values from the caches, so this is coherent), the whole conditional layer can go: `_should_propagate`, create_action's None return, the ~26 `Optional[ListAction]` result fields + None-checks across handlers, and the `ListFighter.clone()` recompute branch. I'd do it as a follow-up rather than in this PR — it touches every handler result type — and ideally move the bootstrap into the model/manager layer first so the invariant is structural. Happy to file the issue.

tgvashworth: Filed as #1980.

PR #2018, opened 2026-07-21
feat: data-driven promotions in the advancement flow (promotions Phase 2)

### LINKED ISSUE #1467: Promote to Champion advancement missing for some Prospect fighter types
## Summary

The "Promote to Champion" advancement is not appearing for some Prospect fighter types. While the promotion advancement works for certain fighters, it appears to be missing or not fully implemented for Prospects that can promote into two different types of Champion.

## Details

- A user reported that promotion advancements were not visible for their Prospect fighters.
- Investigation confirmed that the feature works for some fighter types but not others.
- The suspected root cause is that **some Prospects can promote into two different types of Champion**, and this dual-promotion path may not have been fully implemented.

## Context

- Previous work on promotion advancements was done in PR #1362 (closing #1150), which updated `core/views/fighter/advancements` to account for the `skill_promote_champion` advancement.
- The promotion flow works correctly for fighter types with a single Champion promotion path.
- The gap appears to be specifically for Prospects with multiple possible Champion promotion targets.

## Expected Behavior

All Prospect fighter types should see the "Promote to Champion" advancement option, including those that can promote into more than one type of Champion. In the case of multiple promotion paths, the user should be able to select which Champion type to promote into.

## Raised By

Reported by **manofwaaagh** during discussion with **tom_gyrinx**.

---

### Source

This issue was created from a Discord conversation flagged by tom_gyrinx.

---

**Source:** [Discord message](https://discord.com/channels/1337524316987985963/1408118168060432514/1472246817927205018) | Requested by: tom_gyrinx

### LINKED ISSUE #1596: Promote to Specialist option missing for Juves
See also #1467 — Juves don’t appear to currently have promotion options (promote to specialist) available. Under current system they need to be manually added (by Tom), possibly this is an indication that we need to change backend selection options to allow content-admin to specific advancements for fighters.

Example here:

Image

tgvashworth:
## Summary

- Wires the advancement wizard through `ContentPromotionPath`: promotions are now offered, applied, and reversed from content rows instead of hardcoded Python. **This closes #1596** — an admin authors a Juve→Specialist path and Juves see it immediately, no code change.
- Fixes the roll-12 latent bug: the rulebook's Ganger 2D6 table promotes on a roll of 2 **or** 12; the old code only implemented 2, so a rolled 12 silently offered Willpower instead. The prefill is now driven by the path's `rolls` list.
- RAW-faithful by construction: applying a promotion relabels the fighter's category and adds the flat `cost_increase` — statline and base cost are untouched.

## How it works

- New `ADVANCEMENT_PROMOTION` advancement type with a `promotion_path` FK (migration 0182). Skill-less promotions (most house Juve/Prospect paths) go type → confirm; promotions that bundle a skill (the core Ganger/Specialist paths) ride the existing skill-selection steps and store as promotions.
- Reversal is rank-driven: deleting a promotion falls back to the highest-ranked promotion the fighter still holds, across both eras — data rows via their path's `rank`, historical `skill_promote_*` rows via a static map that works even with no DB rows.
- **Multi-target paths (Prospect → Forge Boss *or* Stimmer) are deliberately not offered yet** — that needs the target-selection step (next phase, closes #1467). They're authored and waiting; the wizard gates them so the player's "which type?" choice can't be silently dropped.
- Legacy compatibility: stored `skill_promote_specialist`/`skill_promote_champion` rows keep applying, displaying, and reversing forever; old mid-flow URLs stay valid.

## Test plan

- [x] New `test_promotion_paths.py` (8 tests): Juve sees its authored path; full wizard flow applies it as a label-only relabel (base cost asserted unchanged, flat bump only); roll 2 **and** 12 prefill the promotion; stat rolls unaffected; rank-driven and mixed-era reversal; multi-target gating.
- [x] Full suite: **3,548 passed** (1 perf-snapshot fixture updated — 2 queries gain the new FK column, query *count* unchanged, no N+1).
- [x] Manual re-run of the earlier browser-flow scenarios in the dev DB: the Juve completed her authored promotion through the real wizard; a rolled 12 now prefills Promote to Specialist (+20¢); the single-target Prospect path is offered; the two-target Prospect correctly sees nothing yet.

## Next (not in this PR)

- Target-selection step + the access-only promotion pointer on `ListFighter` → closes #1467.

Closes #1596
Refs #1467


tgvashworth:
Superseded by #2019, which now contains this diff plus Phases 3–5 against main — un-stacking per maintainer preference. The Copilot review feedback here is being applied on #2019.

PR #2040, opened 2026-07-24
refactor: build the cost-change signal pairs from one factory

tgvashworth:
Redo of #1909, which was ~5 weeks stale and unrebasable — `signal_handlers.py` had gained +419/−116 lines from nine other PRs underneath it. Same idea, written fresh against current main.

## What changed

Ten price-bearing content models each had a hand-written `pre_save` cost detector and a hand-written `post_save` fan-out — **twenty near-identical receivers**. They're now one `register_cost_change()` factory plus ten registrations. **No behaviour change**; `signal_handlers.py` loses ~260 lines.

Details worth knowing:

- **All 20 `dispatch_uid`s are passed explicitly and preserved byte-identical.** They predate any single scheme (`ContentFighter` is `content_fighter_base_cost_change` but `content_fighter_cost_action`), and `dispatch_uid` is the identity Django dedupes registrations on — so they can't be derived from the model. A test asserts every one is registered *and live*, and that the registered set matches the expected set.
- **The two "special" detectors turned out to be the same shape.** Both do a raw `get_old_field` compare on top of the int compare — the weapon accessory for `cost_expression`, the expansion item to catch a `0 <-> None` edit that the int coercion hides (None means "use base price", 0 means "free"). One optional `also_watch` parameter covers both, with each rationale kept at its call site.
- `sync_auto_equipment_cost` stays hand-written — it's a genuinely different `post_save` on `ContentFighter`, including its `raw=` fixture-loading guard from #1995.

## The part that isn't obvious: receiver lifetime

Django holds receivers **weakly** by default. The module-level functions this replaces each had an implicit strong reference (their own module global); a closure built inside a factory does not — so it gets collected and the signal **silently stops firing**, with no error.

The trap is that **this cannot reproduce in dev or CI**: `Signal.connect` only calls `func_accepts_kwargs(receiver)` when `settings.DEBUG` is true, and that check runs through an `@lru_cache` keyed on the function — which pins it. With DEBUG on every receiver is held alive incidentally; with DEBUG off (production) nothing holds it. Green locally, dead in prod.

Both receivers therefore connect with **`weak=False`**, which puts the guarantee at the connect site — the same idiom already used at `core/models/state_machine.py:135`. (An earlier revision kept a module-level strong-ref list instead; review rightly pointed out that a list is a step decoupled from the `@receiver` calls, so anyone adding a third receiver and forgetting to extend it would silently reintroduce the bug.)

The regression test runs with `DEBUG=False` and first asserts that an ordinary weakly-connected receiver really does die, so it can't pass vacuously.

## Test plan

- [x] Full suite: **3643 passed, 13 skipped**.
- [x] Cost-signal suites (`test_content_cost_signals`, `test_propagate_content_cost_change`, `test_cost_propagation`, `test_list_refresh_cost`).
- [x] Verified out of band, with DEBUG forced off *before* app loading: all 20 receivers live after a forced GC, all 20 dispatch_uids and all 20 `@traced` span names byte-identical to main, per-receiver closure captures correct, and `ContentFighter`'s post_save ordering unchanged (cost action still before `sync_auto_equipment_cost`).
- [x] Mutation-tested: dropping `weak=False` fails the regression test.
- [x] `ruff` clean; no migrations.

## Known pre-existing gap (not addressed here)

None of these ten receivers has a `raw` guard, so during `loaddata` / `loaddata_overwrite` every price row whose cost differs from the local row calls `set_dirty()` — bulk-writing to `ListFighterEquipmentAssignment` — and enqueues a propagation task on commit. The #1991/#1995 fixture-load sweep missed this pair. `main` has the identical gap, so fixing it here would break this PR's no-behaviour-change property — but the factory turns it from ten edits into one.

Refs #1863, #1855

### DIFF
--- a/gyrinx/content/models/signal_handlers.py
+++ b/gyrinx/content/models/signal_handlers.py
@@ -38,270 +38,147 @@ logger = logging.getLogger(__name__)
 
 
 # =============================================================================
-# Pre-save signal handlers
+# Cost-change signals
 #
-# These signals detect when cost fields change on content models and mark
-# affected core objects (assignments, fighters, lists) as dirty.
+# Every price-bearing content model needs the same pair of receivers: a
+# pre_save that spots a cost edit, flags the instance and marks affected core
+# rows dirty; and a post_save that enqueues propagation (audit actions and
+# cache deltas). Rather than hand-write that pair ten times, they are built by
+# one factory and registered per model below.
 # =============================================================================
 
 
-@receiver(
-    pre_save, sender=ContentEquipment, dispatch_uid="content_equipment_cost_change"
+# Django stores receivers as weakrefs, so a receiver needs a strong reference
+# somewhere or it is collected and the signal silently stops firing. The
+# module-level functions this factory replaced had one implicitly (their own
+# module global); closures built inside a factory do not, so we keep them here.
+#
+# This is easy to get wrong because it does not reproduce in development or in
+# the test suite: `Signal.connect` only calls `func_accepts_kwargs(receiver)`
+# when `settings.DEBUG` is true, and that check runs through an `lru_cache`
+# keyed on the function — which pins it. With DEBUG on, every receiver is held
+# alive incidentally; with DEBUG off (production) nothing holds it, and the
+# receiver is collected. Green locally, dead in prod, no error either way.
+_COST_CHANGE_RECEIVERS = []
+
+
+def register_cost_change(
+    model, *, change_uid, action_uid, field="cost", also_watch=None
+):
+    """Register the standard cost-change receiver pair for a content model.
+
+    Args:
+        model: the content model class whose price edits should propagate.
+        change_uid: ``dispatch_uid`` for the ``pre_save`` receiver.
+        action_uid: ``dispatch_uid`` for the ``post_save`` receiver. Both are
+            passed explicitly rather than derived from the model, because the
+            shipped strings predate any single scheme — ``ContentFighter`` is
+            ``content_fighter_base_cost_change`` but ``content_fighter_cost_action``
+            — and ``dispatch_uid`` is the identity Django dedupes registrations
+            on, so they must stay byte-identical.
+        field: the cost field to compare, through the int-coercing helpers.
+        also_watch: optional second field, compared by its *raw* stored value,
+            whose change also counts as a cost change. Two models need this,
+            for different reasons — see the call sites.
+    """
+
+    @receiver(pre_save, sender=model, dispatch_uid=change_uid)
+    @traced(f"signal_{change_uid}")
+    def on_cost_change(sender, instance, **kwargs):
+        old_cost = get_old_cost(sender, instance, field)
+        if old_cost is None:
+            return  # New instance, nothing existing to reprice
+
+        changed = old_cost != get_new_cost(instance, field)
+        if not changed and also_watch is not None:
+            old_raw = get_old_field(sender, instance, also_watch)
+            new_raw = getattr(instance, also_watch)
+            changed = old_raw is not MISSING and old_raw != new_raw
+
+        if changed:
+            instance._cost_changed = True  # Flag for post_save to create actions
+            # The pre-change value rides to the async task: amount-snapshot
+            # (DERIVED) receipts are maintained by delta, which needs it.
+            instance._old_cost_for_propagation = old_cost
+            instance.set_dirty()
+
+    @receiver(post_save, sender=model, dispatch_uid=action_uid)
+    @traced(f"signal_{action_uid}")
+    def on_cost_change_saved(sender, instance, created, **kwargs):
+        if created or not getattr(instance, "_cost_changed", False):
+            return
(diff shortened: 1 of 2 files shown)


tgvashworth:
Review round applied (`53a4c7ec`). An independent pass verified the no-behaviour-change claim programmatically rather than by eye — introspecting the live registry under both this branch and main, with DEBUG forced off before app loading. It confirmed all 20 dispatch_uids and all 20 `@traced` span names are byte-identical, the per-receiver closure captures are correct (only `ContentFighter` has `field='base_cost'`, only the two expected models have `also_watch`), and — a check I hadn't thought to make — that `ContentFighter`'s post_save **ordering** is preserved, with `content_fighter_cost_action` still firing before `sync_auto_equipment_cost`.

**Changed: `weak=False` replaces the manual strong-ref list.** This is a better guard and I've taken it. The list was a step decoupled from the `@receiver` calls, so a maintainer adding a third receiver inside the factory and forgetting to extend it would silently reintroduce the prod-only failure the guard exists to prevent. `weak=False` puts the guarantee at the connect site; `_COST_CHANGE_RECEIVERS` and its test are gone. The codebase already uses this idiom at `core/models/state_machine.py:135`.

Two test bugs fell out of that, both worth flagging:

- The liveness helper did `entry[1]()` to dereference a weakref. Under `weak=False` that entry **is** the receiver — so it would have *invoked the receiver* with no arguments. Now type-checked first.
- The dispatch_uid test asserted membership, not liveness. A dead weakref lingers in `signal.receivers` until Django sweeps it, so that test would have passed even if every shipped receiver were dead. It now asserts liveness and compares the registered set against `COST_CHANGE_PAIRS`, so an unlisted eleventh model fails there rather than somewhere less obvious.

Verified: with DEBUG forced off before app loading, all 20 receivers are live after a forced GC; dropping `weak=False` fails the regression test. Full suite 3643 passed, 13 skipped. Module docstring refreshed.

**Not bundled here — a real pre-existing bug worth its own issue.** None of these ten receivers has a `raw` guard, so during `loaddata` / `loaddata_overwrite` (the prod content-sync path) every price row whose cost differs from the local row calls `set_dirty()` — bulk-writing to `ListFighterEquipmentAssignment` — and enqueues a `propagate_content_cost_change` task on commit. The #1991/#1995 fixture-load sweep added `if kwargs.get("raw"): return` to the core receivers and to `sync_auto_equipment_cost` but missed this pair. `main` has the identical gap so it isn't a regression, and fixing it here would break this PR's no-behaviour-change property — but the factory has just turned it from ten edits into one.

tgvashworth:
Description updated to match the implementation — it still described the earlier module-level strong-ref list, which `weak=False` replaced in `53a4c7ec`. Good catch.

tgvashworth:
Fixed in the latest push — scoped the registration-drift assertion to the `_cost_change` suffix instead of the `content_` prefix, so an unrelated `pre_save` receiver on a content model can't fail it. Verified by mutation that it still catches an unlisted registration.

tgvashworth:
Applied — each generated receiver now takes its `__name__`/`__qualname__` from its dispatch_uid, so introspection can tell the ten models apart (e.g. `pre_save` on ContentFighter now reports `content_fighter_base_cost_change` rather than a tenth anonymous `on_cost_change`). Noted in the comment that raw traceback frames still show the shared code-object name — this fixes what logging and error reporting actually read. Full suite 3644 passed.

tgvashworth:
Applied in `5de09bcb` — the drift check now filters by liveness via the existing `_live` helper, so a lingering dead weakref can't be counted as an extra registration. Verified it still catches a drifted/unlisted uid.

tgvashworth:
Two findings, one applied and one declined.

**Applied** — mirrored the drift assertion for `post_save` action uids. You're right that the exact-set check only covered `pre_save`, so a hand-written or unlisted `_cost_action` receiver would have passed unnoticed.

**Declined — receiver ordering is unchanged.** The concern was that the generated `post_save` handlers now connect before `sync_auto_equipment_cost`. They did on `main` too:

- `main`, `signal_handlers.py:790` — `@receiver(post_save, sender=ContentFighter, dispatch_uid="content_fighter_cost_action")`
- `main`, `signal_handlers.py:801-803` — `sync_auto_equipment_cost`

So the cost action was already registered first, and the refactor preserves that. Confirmed at runtime rather than by reading — `post_save._live_receivers(ContentFighter)` returns `['post_save', 'content_fighter_cost_action', 'sync_auto_equipment_cost']`, matching main. The independent review pass checked the same thing separately.

(The orphan-pin handlers are `pre_delete`, a different signal, so there's no ordering relationship with these.)

tgvashworth:
Applied — good catch, the per-model assertions weren't actually per-model. `_live()` matched on `dispatch_uid` alone, so a uid wired to the wrong sender would still have satisfied that model's check. The lookup key is `(dispatch_uid, sender_id)`, so it now matches the sender too. Verified by mutation: pointing a registration at the wrong model fails, where previously it passed.

PR #2071, opened 2026-07-26
feat: gang plate on classic sheets, XP on web cards, and Lore & Notes cards

### LINKED ISSUE #1816: Expanded print options (notes, XP and resources)
It would be useful to expand our print options to include some other things people may wish to see on printouts:

1. Showing fighter notes (this may require some kind of expanded card view)
2. Fighter XP (shown on card)
3. Gang resources (shown as panel that appears like a card)

Number 1 is the trickiest so we can discuss separately if needed.

tgvashworth:
Expands what a gang printout can include, and closes a gap between the two card styles that had gone unnoticed.

## What changed

**Classic sheets now carry gang-level information.** Printing in the "classic" grimdark style rendered fighter cards and nothing else — no assets, resources, attributes or stash. Those toggles were in the print-config form but did nothing when classic was selected, with no indication why. Classic sheets now open with a *gang plate*: same 100x110mm card, carrying the gang's name, house, rating and credits, plus its resources, assets, attributes and stash contents. It respects the same `include_*` toggles as the web sheet, so a config that hides assets hides them on both, and the plate is skipped entirely when there's nothing to put on it.

**XP can print on web fighter cards.** The classic card has always printed a fighter's XP; the web card explicitly suppressed it in print. A new "Show XP on Fighter cards" option (on by default) brings the web sheet into line.

**Lore and notes can print two ways.** Fighter cards have no room for free-form prose, so lore and notes get their own cards.

- As a standalone printout at `list/<id>/print/lore-notes`, combining what the Lore and Notes pages show separately, in either card style. Linked from both of those pages.
- Or as part of the main printout: the new "Include Lore & Notes Cards" config option (off by default) tiles a note card for the gang and for each fighter who has lore or notes in among the fighter cards, so both come off the printer in one pass. It follows the config's fighter selection, so a fighter left off the sheet has no note card either.

Fighters with nothing written about them are skipped rather than printing "No lore added yet", and private notes print only for the gang's owner. On the classic style, each of Lore and Notes gets a translucent write-in box beneath it, sharing whatever height the prose leaves, so there is always room to add to them by hand at the table.

Under the hood, the two new plate types share the fighter card's plate, nameplate and cog, but flow labelled text where the fighter card has its fixed statline and weapons regions. Both are previewable in the Print Lab (`/_debug/print-lab/`) as presets, and the lab gained two real-data sources — the gang plate now appears in its "whole gang" view, and a new "lore & notes" source renders those plates for a real gang.

## Test plan

- [x] `pytest -n auto` — 4007 passed, 13 skipped
- [x] New tests in `gyrinx/core/tests/test_print_expanded_options.py` cover the gang plate's sections and toggles, XP on/off, the write-in boxes, the standalone lore/notes printout in both styles, and the combined sheet (both styles, fighter selection, owner-only private notes)
- [x] Checked visually against a real gang: classic sheet with the gang plate, both combined sheets, and lore plates against sparse, ordinary and deliberately overlong prose
- [x] `./scripts/fmt.sh`, `npm run css-lint` (no new errors), raw-markup and cotton gates pass
- [ ] Worth a look on paper — the classic layout is mm-precise and hasn't been through a real printer

## Notes

`include_xp` defaults to on, so existing saved print configs will start showing XP on their printouts — that was the requested default. `include_lore_notes` defaults to off, so nothing else about existing configs changes.

Closes #1816


tgvashworth:
## Code review

Found 4 issues, all fixed in 2ff3a1a.

1. Rich text was flattened with `strip_tags` alone, which runs paragraphs together (`<p>One.</p><p>Two.</p>` becomes "One.Two.") and leaves entities encoded, so an ampersand typed in the editor printed as `&amp;` on the plate. Also affected the classic fighter card's Notes region. (bug)

https://github.com/gyrinx-app/gyrinx/blob/82e9bc2dfbbf6a4e56e0d0bd10cb0f2e0eba53e5/gyrinx/core/print_cards.py#L563-L576

2. `include_xp` did not govern the classic card, which fills its own XP region — so unticking "Show XP on Fighter cards" hid XP on the web sheet only, contradicting the label. (bug)

https://github.com/gyrinx-app/gyrinx/blob/82e9bc2dfbbf6a4e56e0d0bd10cb0f2e0eba53e5/gyrinx/core/views/list/views.py#L728-L740

3. The two new checkboxes were hand-written `form-check` markup (gyrinx/core/templates/CLAUDE.md says: "label + widget + help + errors by hand | `<c-form.field :field="form.x" />`", and CLAUDE.md says: "Adding raw markup for a pattern a component already covers is a regression."). Converting also fixed a missing `aria-describedby` link and a double-escaped ampersand in the label.

https://github.com/gyrinx-app/gyrinx/blob/82e9bc2dfbbf6a4e56e0d0bd10cb0f2e0eba53e5/gyrinx/core/templates/core/print_config/form.html#L62-L72

4. The test's stash fighter was built with `ContentFighter.objects.create` / `ListFighter.objects.create` (CLAUDE.md says: "Do not manually create users, houses, fighters, campaigns, or lists inline when a fixture or factory already exists.").

https://github.com/gyrinx-app/gyrinx/blob/82e9bc2dfbbf6a4e56e0d0bd10cb0f2e0eba53e5/gyrinx/core/tests/test_print_expanded_options.py#L101-L118

CodeRabbit's three points: 1 and 3 above are its rich-text and XP-label findings, taken as bugs rather than doc tweaks. Its SCSS comment-spacing point is applied, though prettier and stylelint both pass either way.

Not changed, flagged for the record: a classic sheet for a gang with no resources, assets, attributes or stash gets no gang plate at all, so its name and rating appear nowhere (the classic sheet has no header). That mirrors the web sheet, where the assets card only appears when there are assets, but it is a deliberate choice worth revisiting if printed classic sheets turn out to need identifying.

PR #2138, opened 2026-08-07
chore: target ruff at 3.14 and adopt the syntax

tgvashworth:
chore: target ruff at 3.14 and adopt the syntax

Follow-up to #2136, which moved the runtime to 3.14 but deliberately pinned
ruff's `target-version` behind it at py312. This removes that pin.

## ⚠️ Do not merge until 3.14 is deployed and has settled

The pin existed for one reason: with ruff targeting the real runtime, the
formatter emits PEP 758 `except A, B:` — syntax that parses on 3.14 and nowhere
else. Merging this closes the option of rolling the Docker base image back to
3.12, because the code would no longer parse there.

That is not a theoretical claim. Compiling this branch with 3.12 fails on **10
files in `gyrinx/` alone**. Once this is in, 3.14 is the only runtime that works.

So the ordering matters: #2136 first, deployed, and given long enough in
production to trust. Then this.

## What changed

- `target-version = "py314"`, matching `requires-python`.
- 56 findings that only appear at the 3.14 target, all auto-fixed — 51 of them
  `UP037` (quoted annotations, unnecessary under 3.14's annotation semantics),
  the rest PEP 604 unions and default type args.
- The formatter rewrote 42 `except (A, B):` into `except A, B:` across 32 files.

44 files, 103 insertions, 117 deletions. All mechanical; no logic touched.

## Verification

- `ruff check` clean, `ruff format --check` stable at 882 files.
- `pytest`: **4194 passed, 13 skipped** — unchanged.
- djlint, prettier, migration conflicts, bandit --baseline: all clean.
- The production image builds on `python:3.14.6-slim`, loads all 34 apps in the
  container, passes `manage check`, and imports the WSGI callable.
- And deliberately confirmed in the other direction: the branch no longer
  compiles on 3.12. That is the intended consequence, stated here so it is not a
  surprise later.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01ToFTcKVBQptbR7eTWkWuWH


tgvashworth:
**Holding this one deliberately — do not merge yet.**

It should land only after #2136 has deployed and 3.14 has run in production long enough to trust, because merging it removes the option of going back. Verified rather than assumed: this branch fails to compile on 3.12 in 10 files under `gyrinx/` alone, so once it is in, rolling the Docker base image back would leave code that does not parse.

Worth recording one thing found while building it, which was not in the original plan. Adopting PEP 758 broke **pre-commit itself**: several hooks parse Python with Python (`check-ast`, `debug-statements`, bandit) and were building 3.12 environments, so every commit failed with a SyntaxError that read like a bug in the code rather than in the tooling. Fixed here by pinning `default_language_version` to python3.14 — but it is a good illustration that a syntax-level upgrade reaches past the application into the toolchain around it.

PR #2160, opened 2026-08-08
feat: ingest the reworked sheets, and an Ingest page to run it from

tgvashworth:
The pre-ingest spreadsheets were reworked and the change was semantic, not
cosmetic, so the importer has been rebuilt around what they now say — and given
a page, because until now `plan_ingest` and `perform` were reachable from the
test suite and a shell and nowhere else.

## The importer

**What changed in the sheets.** Equipment is now a single catalogue: one row per
thing the game sells, typed by an `Assignable` column, and the only place a
reference price lives. Weapon profiles carries statlines and nothing else.
Equipment lists names a collection per `Title`. Most importantly they share a
computed `ID` — `Name (Profile) (Category ← Section)` — which reproduces exactly
from its four parts across all three sheets.

**That ID is why this is a rewrite.** Resolution used to go by name, and name is
not an identity: eight weapons share a printed name across categories — a power
fist is Exo kit *and*, separately, a Power weapon at a different price.
Name-matching silently merged them. Everything now resolves through the parsed
ID, and where two catalogue rows genuinely share a name both are kept, told
apart by the author-facing qualifier.

**Prices got simpler**, because the sheets keep one rule: a thing is either sold
at the Trading Post at a printed price, or it is equipment-list only — `Cost "-"`
and `TP "E"` always appear together. The old two-pass price backfill is gone. An
entry overrides only where its list actually disagrees, which across the real
sheets is four entries out of 655.

**Nothing builds a Trading Post**, because nothing needs to: membership is having
a trade point price, swept in by standard content. Ingest sets the field and the
post fills itself — verified at 94 items.

**Grenades are weapons that take no slot.** The sheets type them Wargear because
they don't count against the weapons a fighter holds, but they have firing lines,
and wargear carries no statline. A Wargear row with a statline is now a Weapon
with `slots = 0`, which records the fact the typing stood in for. Ten rows.

**Fighters, and two model gaps they exposed.** A rule's annotation is now part of
its identity, as a trait's already was — a leash at four distances was four rules
sharing a name, and only one could exist. And a specialisation can now say who may
use a thing: `(Gunner specialist only)` is an arm of usable-by like a subtype, so
those restrictions became real instead of notes. The eight specialisations from
the rulebook's Specialist table are standard content, each wired to the skill it
grants.

## The page

`/n26/authoring/ingest/`, staff-only. Upload the sheets; two buttons over the same
files. Preview plans and writes nothing, Import does what Preview showed — they can
be two buttons rather than a wizard because planning is pure, so nothing is carried
between them and an abandoned upload leaves nothing behind.

Problems are grouped by shape rather than listed by line: forty problems are five
kinds of problem, and each shape keeps a few real line numbers under it.

**Severity now has a stated rule.** An error would write something *wrong*, or names
a state that is never legitimate. A note means the upload writes *less* than the
sheet asked for and says which lines. Most shortfalls are notes — refusing a
thousand good rows over four the sheets aren't ready for helps nobody.

**A danger zone** deletes everything an import wrote and leaves the foundations
standing, on its own confirmation page that counts what would go first. Two
contracts, both tested: after a clear every standard-content seed still reports
complete, and import → clear → import lands on identical counts. Counting and
deleting read one definition, so the confirmation cannot promise different numbers
from the ones that go.

## How to try it

```
/n26/authoring/ingest/
```

The four CSVs go in the four fields. Against the current sheets that previews
cleanly with **2 errors and 39 notes**; both errors are genuine mistakes in the
spreadsheet (a mistyped catalogue row, and two fighters sharing one qualifier),
and the rest is the team's fix list. Fix those two cells and all four sheets
import: 145 fighter profiles, 267 weapons, 67 wargear, 19 named equipment lists,
655 entries — and re-running creates nothing.

## Notes for review

- Two migrations, both additive: `library/0022` widens a rule's uniqueness to
  include its annotation (a constraint can only get weaker, so no existing row can
  violate it), `library/0023` adds the usable-by-specialisation M2M.
- The Trading Post listing costs two more queries — one prefetch per sweep — which
  is the price of a fourth usable-by arm and much cheaper than the per-item queries
  omitting it would cause. The budget test carries the new number.
- `n26` suite: 1335 passing.


tgvashworth:
Four of five taken, in 1af8344a. Thanks — the first two were real and one of them crashed.

**WeaponProfile / Weapon resolution ignoring the ID's category** — correct, and it was the more serious of the two. Resolution against the pack now uses the whole ID, falling back to a bare name only where the pack holds exactly one row of that name. That keeps a hand-authored item filed under its own category findable without ever guessing between two "Power fist" rows. Reproduced before fixing: a list-only upload resolved to the 0-credit Power weapons row instead of the 105-credit Exo one.

**Dropped fields on the "exists" path** — correct, and it was not just a mis-detect. `_entry_exists` reads `fields["weapon"]` for a profile, so uploading a list of ammo against an existing catalogue raised `KeyError: 'weapon'` outright. The qualifier and the weapon key are both carried now, and that reader takes the key defensively too.

**`clear_imported(pack=...)` ignoring pack for modifiers** — correct. Latent, since the page only ever calls it unscoped, but it is the function's contract and it now scopes like everything around it.

**Test upload without a filename** — taken; `SimpleUploadedFile` with the name the helper was already being handed and throwing away.

**Cost/TP validation** — declining, because the premise does not hold. The two fields are read independently and both honoured: `Cost=45, TP=E` yields price 45, exclusive, no trade-point price, which is a coherent item — 45 credits, never sold at the Trading Post. Nothing is coerced and nothing downstream depends on the pairing. The "always appear together" in the docstring describes what the current sheets happen to contain, not an invariant the code enforces, so blocking on it would refuse a row the model can represent perfectly well.

PR #2161, opened 2026-08-08
feat: ingest weapon accessories

tgvashworth:
The sheets now type a sight or a suspensor `Weapon Accessory`, which the
importer did not know. Those rows were landing as wargear — the wrong table,
and wrong on a card: an accessory bolts onto a weapon rather than being
carried, and the library has always had its own kind for it.

An accessory plans like wargear — a name, a home, a price — so the two share
one branch rather than getting one apiece, and it resolves, performs and
clears alongside everything else.

**What it fits is not in the sheets.** `fits_category` ("Las Weapons Only")
and `fits_asterisked` have no column, so an imported accessory fits anything
and is narrowed by hand afterwards. That is the honest reading of a column
that does not exist, rather than a guess.

## Two things the change dragged in

**Creating one could not say it was Trading Post exclusive** — `is_exclusive`
was missing from `create_weapon_accessory`, the same gap wargear had. Added.

**The Trading Post would have quietly lost them.** The post sweeps by kind, and
its two sweeps were weapons and wargear — so retyping seven accessories *out*
of wargear takes them off the shelf they are sold from. Caught by running the
retyping against the real export: sweeps dropped 94 → 87, exactly the seven.
The post now sweeps accessories too.

That seed also had to learn to top up rather than skip: it previously created
the post only when absent, so an existing post would never gain the new sweep.
It now adds any missing sweep, reports `incomplete` until it has them all — so
the Foundations page is how anyone finds out — and stays idempotent. Verified
against a real database: two selectors and `incomplete` before, three and
`complete` after, still three on a second run.

## How to try it

Against the current export with the eight accessory rows retyped:

```
errors: 0
WeaponAccessory: 7        (the eighth is the duplicate Mono-sight, noted)
wargear: 67 → 60
trading post sweeps: 94   (unchanged, once the seed is topped up)
```

`Mono-sight` and `Telescopic sight` land on 7 lists each, `Infra-sight` on 6.
A clear takes them with everything else.

## Notes for review

- No migrations. `WeaponAccessory` already existed, is already an entry kind on
  `CollectionEntry`, and is already sweepable.
- The Trading Post query budget moves 13 → 14: one query for the third sweep.
  An accessory is not `UsableBy`, so it adds no restriction prefetches. The
  budget test carries the new number and says why.
- `n26` suite: 1340 passing.

### DIFF
--- a/n26/library/authoring.py
+++ b/n26/library/authoring.py
@@ -219,6 +219,7 @@ def create_weapon_accessory(
     name,
     price=0,
     trade_point_price=None,
+    is_exclusive=False,
     category=None,
     fits_category=None,
     fits_asterisked=False,
@@ -235,6 +236,7 @@ def create_weapon_accessory(
         name=name,
         price=price,
         trade_point_price=trade_point_price,
+        is_exclusive=is_exclusive,
         category=category,
         fits_category=fits_category,
         fits_asterisked=fits_asterisked,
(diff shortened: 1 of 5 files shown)


tgvashworth:
Taken in 9b769508 — correct, and worse than described: the same lookup-order problem also picked between two rows of one kind told apart by a qualifier. Every kind is now asked at once and every row of each counted; one match resolves, anything else is refused and said. Three tests: one kind, two kinds sharing a name, two rows of one kind.

PR #2168, opened 2026-08-11
perf: the print page builds the gang once, and kitted lines no longer cost a query each

tgvashworth:
Two query fixes on n26 pages, found by profiling against a content-rich database.

- **The print page built the whole gang twice.** Printing derived the gang once to fill in the header and stash blocks, then derived it all over again — rows, statlines, modifier index — to lay out the cards. A gang card now keeps the flat rows its members' cards were dealt from, so a print config's ticked weapon selection re-deals those cards in memory instead of refetching, and one build serves the header, the stash and every card. The print view no longer calls `render_gang` at all: the template reads the gang's own figures, and the stash block is derived from the same card (`stash_lines` in `n26/core/render.py`). On a full roster the page drops from 84 queries to 47 and roughly halves its SQL time (81ms → 53ms); smaller gangs halve too.
- **A line that comes with kit cost a query to price.** An equipment line's advertised price composes in the price of the built-in set it ships with — a mount and its saddle, say — but neither a curated listing (`n26/core/browse.py`) nor a selector sweep (`n26/library/models/collection.py`) fetched that set, so every kitted line on a page paid a query of its own at the till. That is only two or three lines per equip page today, but it grows one-per-line as kitted content lands. Both paths now load the set with the listing.

Nothing changes on screen or on paper: same cards, same prices.

**Test plan**

- [x] Full suite green in CI
- [x] New guard in `n26/core/test_views_print.py`: the print page's query count stays flat as models and gear are added to the roster
- [x] New guard in `n26/core/test_views_print.py`: the page fetches assignment rows exactly twice — the gang's own and its stash's — which is what "derived once" looks like in queries
- [x] New guard in `n26/tests/sandbox/test_collections.py`: a listing of twelve kitted lines costs the same queries as a listing of two, and each package still prices correctly. Against the previous code this measured 12 vs 22.

Sheet and hire pages were profiled too and are already at their fixed budgets — their repeated query shapes are parallel prefetch paths sharing a table, not duplicate work.

**To try:** open a gang and hit Print, then print a saved config with a subset of models and weapons ticked. Both should look exactly as they did before.

### DIFF
--- a/n26/core/browse.py
+++ b/n26/core/browse.py
@@ -241,6 +241,10 @@ def browse(collection, terms=EQUIPMENT_LIST):
         collection.entries.prefetch_related(
             *ENTRY_ASSIGNABLE_FIELDS,
             *(f"{name}__category__section" for name in ENTRY_ASSIGNABLE_FIELDS),
+            # What each line always comes with: pricing a line composes
+            # the set's own price in, so a list of mounts prices the
+            # packages without a query per line.
+            *(f"{name}__built_ins" for name in ENTRY_ASSIGNABLE_FIELDS),
             # Use-restriction lists, derived for every kind carrying the
             # mixin — so noting a whole listing costs no extra queries,
             # whichever kind an author narrowed and however they narrowed it.
--- a/n26/core/card.py
+++ b/n26/core/card.py
@@ -222,12 +222,34 @@ class GangCard:
     #: both card kinds are computed by one function, and one that read
     #: its own grants on one card and not the other would be a trap.
     granted: list[Node] = field(default_factory=list)
+    #: The flat rows each member's card was dealt from, kept so a
+    #: selection can re-deal the cards without another fetch — see
+    #: ``members_under``.
+    member_rows: dict = field(default_factory=dict, repr=False)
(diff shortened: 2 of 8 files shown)


tgvashworth:
Closing: this work is already on `main`. The commit landed as [`a6ffe61b`](https://github.com/gyrinx-app/gyrinx/commit/a6ffe61bdaa60f13f143452c69be3a6c498f133c) — an identical diff, rebased onto a slightly different `n26/core/card.py` base, so `git cherry origin/main origin/n26-perf-2` marks the branch's only commit as already upstream. The PR showed as conflicting simply because its branch tip was stale.

Verified on current `main` (`d65cca2d`): all three query guards from this PR are present and green —

- `test_a_bigger_roster_costs_no_more_queries_to_print`
- `test_the_page_fetches_the_gangs_rows_once`
- `test_lines_that_come_with_kit_cost_no_more_queries_as_the_list_grows`

(`n26/core/test_views_print.py` 18 passed; `n26/tests/sandbox/test_collections.py` query guards 5 passed.)

Nothing left to merge. Branch `n26-perf-2` deleted.

PR #2169, opened 2026-08-11
perf: card builds hydrate rows in narrow passes instead of one wide join

tgvashworth:
Stacked on #2168 — merge that first; this retargets to `main` when the branch goes.

- **The fetch behind every card build was one enormous join.** `card_rows` pulled all eleven assignable kinds and their chains into a single select. Postgres spends roughly 19ms *planning* that select and roughly 1ms executing it, and Django issues no prepared statements, so the planning is paid again on every call — twice per gang build. The design's pinned "two queries" was honest, and the pin was the expensive part.
- **Rows now come back bare and hydrate afterwards.** `_flat_rows` fetches assignments with only their ledger entry joined; `hydrate_rows` then loads everything a build reads through one `prefetch_related_objects` pass per relation. Each pass is a single-table select that plans in microseconds, a kind no row names skips the database entirely, and a build's row sets (a model's plus its gang's, or a gang's plus its stash's) hydrate together so the passes are paid once. It is the same trade the modifier index and the browse listings already made; the card build was the last wide join standing.

**Effect:** `build_gang_card` drops from ~42ms to ~14ms (best of five, content-rich database). Sheet and print pages lose 30–70% of their SQL time, most visibly on small and medium gangs — a small gang's sheet goes from 62ms to 17ms of SQL. Query counts rise by a fixed ~12 narrow fetches per page, and total time falls everywhere cards are built.

One deliberate consequence: stash rows used to be fetched without statlines, since nothing reads a statline off storage. They now ride the shared hydration pass with the rest of the gang's rows, because a separate pass for a handful of rows would repeat every narrow query for them.

**Test plan**

- [x] Full suite green in CI
- [x] The tests that pin exact query counts re-pinned to the new budget — two row queries plus a fixed hydration — with the decomposition written into their docstrings
- [x] The two grow-and-compare tests now seed a weapon before their first measurement: hydration only queries for kinds actually present, so introducing a new kind mid-comparison would have measured the kind arriving rather than the scaling
- [x] Design language updated wherever it described the old shape — the `n26/core/card.py` module docstring, the `build_card` / `build_gang_card` docstrings, and `n26/core/CLAUDE.md`

**To try:** open a gang sheet and its print page. Nothing changes on screen; they simply spend a fraction of the database time getting there.

### DIFF
--- a/n26/core/CLAUDE.md
+++ b/n26/core/CLAUDE.md
@@ -6,8 +6,8 @@ stage has one job:
 ```
 models/        rows: Gang, Miniature, Assignment, LedgerEntry, Stash
 operations.py  the only writer — every change to player data goes through it
-card.py        loads a gang's rows into an in-memory tree, in a pinned
-               two queries
+card.py        loads a gang's rows into an in-memory tree: a pinned two
+               row queries, one shared hydration pass
 effects.py     computes what the rules do to a card — pure, no queries
 render.py      plain dataclasses a template can draw (ModelCard, GangSheet)
 browse.py      collections as one rendered shape, whatever their species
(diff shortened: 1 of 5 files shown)


tgvashworth:
Closing: this work is already on `main`. The commit landed as [`395ccd56`](https://github.com/gyrinx-app/gyrinx/commit/395ccd56) — a byte-identical diff apart from hunk offsets. `git cherry origin/main origin/n26-card-hydration` marks both commits on the branch (this one and #2168's) as already upstream, so there is nothing left to merge.

Its base PR #2168 is being closed for the same reason. Branch `n26-card-hydration` deleted.

PR #2170, opened 2026-08-12
perf: the picker reads its counts from one table, and tab strips prerender their pages

tgvashworth:
Perceived-performance round on the populated collection views — the hire screen's section strip above all, at the volume prod carries (19 sections, ~140 profiles).

**Why a tab press took over a second.** Every count on the picker — 19 tab badges, the section headers, the "N profiles" readout, the narrow-screen section menu — was a getter, and Alpine re-runs getters per read with no caching. Each read walked every registered item through the reactive proxy, and the getters read each other (`visibleSection` → `liveSections` → 19 × `countInSection`), so one press on the strip called `countInSection` **15,733 times** and `matches` 117,715 times: ~1.1s of main-thread work to redraw numbers a tab press cannot change. Measured in Chrome with the real page; the raw compute behind all 19 counts is 2.8ms.

**The fix** is one `x-effect` that walks the items once when they or a filter change and writes every count into a table; tabs, headers, readout and menu read the table. A press recomputes nothing: the reactive flush drops to ~10–40ms (~100–300ms the first time a big section is revealed, once per load), and a real click on the biggest tab now paints without even crossing the event-timing buffer's 104ms threshold. Search and sliders stay live (~50ms a keystroke at this volume). This lands in `<c-n26.collection-picker>` so the equip screen and the gallery get it too.

**The scope tabs** (Gang list / Supplementary / All profiles) are whole server-built pages — the all-profiles hire list prices every card in the library and ships ~4.5MB of HTML. `<c-n26.tab-links>` gains a `prerender` prop emitting [speculation rules](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API) for its non-current tabs; the hire screen asks for `immediate`, so Chrome builds the other scopes in the background and a press is answered from cache. Everywhere else they stay ordinary links. `record()` now ignores requests carrying `Sec-Purpose: prefetch`, so a prerender nobody opens is never an analytics event (tested).

**The gallery** gains a crowded-catalogue specimen under collection-picker — nineteen uneven suppliers, a couple of hundred generated rows, no database — as the page to open whenever a change might make the picker slower.

**To try:** open a well-populated gang → Hire Fighters → All profiles, and press along the section strip; or open `/n26/design/c/collection-picker/` and scroll to "A crowded catalogue". Tab presses should feel instant; scope-tab presses in Chrome should land near-instantly once the page has settled.

### DIFF
--- a/n26/analytics.py
+++ b/n26/analytics.py
@@ -62,7 +62,13 @@ def record(request, noun, verb, obj=None, **context):
     whole roster, an ingest of a whole spreadsheet — is still one event
     carrying counts, because a row-by-row loop turns a page into as many
     writes as it has rows.
+
+    A speculative fetch is not a press. Browsers prefetch and prerender
+    pages nobody has opened yet — tab strips ask for it — and an event
+    recorded then would count readers who never arrived.
     """
+    if request.headers.get("Sec-Purpose", "").startswith("prefetch"):
+        return None
     return log_event(
         user=request.user,
         noun=noun,
--- a/n26/core/hire.py
+++ b/n26/core/hire.py
@@ -44,10 +44,17 @@ class HireOption:
     #: What the hire costs with this option taken and all else default.
     total_price: int
     is_default: bool
-    card: ModelCard
+    #: None when the caller asked for a list without drawn cards — a
+    #: page that fetches each card on demand instead of shipping every
+    #: one. ``preview_model_card`` is the single card such a page serves.
+    card: ModelCard | None
     #: None when the profile offers no alternatives and this is the
     #: synthesised standard one.
     default_set: object = None
+    #: Where a drawn card for exactly this option can be fetched, when
+    #: the surface serves them on demand. A display address, so the view
+    #: that knows its URLs fills it in; empty means the card is inline.
+    card_url: str = ""
 
 
 @dataclass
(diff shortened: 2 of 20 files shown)


tgvashworth:
Pushed e319930d after live testing surfaced a real problem with the eager prerender: a prerendered page is a complete hidden document — DOM, Alpine state, everything — and holding both sibling scopes built behind the visible one meant three copies of the app's heaviest page in memory at once (enough to crash a Chrome tab in dev, where the debug toolbar rides each copy too), plus two unasked-for server builds per visit.

The primitive now separates what to speculate (`speculate="prefetch"` holds bytes; `"prerender"` holds whole pages) from when (`eagerness`, defaulting to intent — hover/touch-down). The hire screen uses **prefetch on intent**: hovering a scope tab pays the server's seconds early and holds only HTML; the press costs just the parse. Prerender remains available for strips of light pages.

tgvashworth:
Pushed ed6b3b33: the cards themselves are now lazy, as discussed. The hire list ships priced rows and each option's card *address*; a new `<c-n26.deferred>` primitive fetches the fragment the first time a row's disclosure opens (placed inside the existing `x-if`, so nothing fetches until asked), with a pulse placeholder, a retry state, and a latch so flicking between options never refetches. `hire_card` serves one option's drawn card — the same derivation the list used to inline — with `Cache-Control: private, max-age=300` since it's library content, identical for every gang. `build_hire_entry(with_cards=False)` skips the effects engine for cards the page won't show; the gallery's sample rows keep their cards inline (no endpoint to fetch from) and the deferred component has its own gallery entry + demo backed by a sample fragment endpoint.

Measured on the all-profiles scope at prod volume: **HTML 4.5MB → 1.0MB, dev server CPU 52s → 2.4s, 64 → 36 queries.** Verified in the browser: expand fetches and Alpine draws the fetched statline; option-specific fragments carry the option's kit (tested); foreign gangs and stray option ids 404 (tested). Full n26 suite green (2554).

tgvashworth:
This got rebase merged in

Issue #2181, opened 2026-08-16
A slot granted by the gang's guest is never asked


tgvashworth:
A modifier on a thing the gang *holds* (a stored pick — Water Guild) can grant a slot to every model, and each fighter's card asks it: the choice row anchors on the gang's assignment riding the member's card (PR #2180).

A modifier on a thing the gang was *given* (a computed guest — a rule an alliance granted, the hidden a charter hands over) can make the same grant, but the guest has no stored assignment on the member's card to hang the choice's address on. The engine deliberately skips recording the row (see the comment in n26/core/effects.py where `given_here` is decided), so the slot is granted and never asked — the same shape as the original Water Guild bug, one hop further down the chain.

The fix is to thread a stable anchor through echoed contributions: the gang-card grant chain roots in a stored gang assignment, which does ride member cards as a broadcast node, so a `Contribution` that carries its chain root's assignment could anchor the choice row and the picks' cause chain. That is a deliberate piece of engine work, not a patch.

No authored content does this yet; the sandbox suites cover the stored-pick case.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DawHQCRuHjqtKFDoZr9kY8

PR #2206, opened 2026-08-16
fix: a question whose slot was taken away is not asked

tgvashworth:
## Summary

- Removing a modifier-granted slot now silences its question: a given slot's choice row stands only while the slot itself does. Before, retraction dropped what a *dead giver* applied, but a slot removed by a still-live giver kept its row — the card asked a question whose subject was gone.
- A new probe suite (`n26/tests/sandbox/test_offer_swap_probe.py`) pins the whole "Subjugator Patrol Officer" story, both the legacy build and the slots-and-picks build.

## Why

The pattern content authors tried and abandoned: a Specialist subtype grants the general specialisation choice; the Subjugator Patrol Officer profile takes it away and grants a narrowed one, so its card should ask the same question over a shorter list. The probe shows the legacy (hidden + offers-a-choice) build computes the right lists but its questions are **anchorless** — a granted hidden is no stored assignment, so the picker has no address and the choice can never be settled, on *any* profile using the shape. That is why it "never worked".

The slots-and-picks build of the same pattern is anchored on the granting subtype's stored assignment and works end to end — narrowed list, settleable pick, reconciled money — once this fix stops the removed general slot's question from lingering beside the narrow one.

## Test plan

- [x] Probe suite: 7 tests — legacy build's lists and plan trace (offer retracted), its anchorless questions pinned as observed behaviour, and the slots-and-picks build asking anchored questions, narrowing per profile, and settling on the bearer
- [x] Full n26 suite: 3,506 passed, 1 xfailed (pre-existing)

Refs #2181 (the echoed-guest anchor gap is the same disease one hop further; this PR fixes the removal side for locally granted slots).

https://claude.ai/code/session_01DawHQCRuHjqtKFDoZr9kY8

### DIFF
--- a/n26/core/effects.py
+++ b/n26/core/effects.py
@@ -856,7 +856,19 @@ def compute(card, index):
         if ModifierIndex.key(contribution.thing) not in dead
     ]
     _fill_choice_slots(computed, offers.items, by_cause)
-    _fill_slot_choices(computed, given_slots.items, by_choice)
+    # A question whose slot was itself taken away is not asked: the
+    # giver may stand, but the given thing is gone — the Subjugator
+    # pattern, where a profile removes the general slot its subtype
+    # grants and grants a narrower one of its own.
+    _fill_slot_choices(
+        computed,
+        [
+            given
+            for given in given_slots.items
+            if ModifierIndex.key(given[0]) not in dead
+        ],
+        by_choice,
+    )
     return computed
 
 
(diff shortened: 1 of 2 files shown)


tgvashworth:
Folded into https://github.com/gyrinx-app/gyrinx/pull/2207. Both commits are rebased onto that branch unchanged; the two fixes are the same machinery seen from opposite ends (a question with no address, and a question whose slot went), so they ship together. The `subjugator-swap` branch is left in place.

Issue #2165, opened 2026-08-08
n26: built-ins added to a profile never reach fighters already hired from it


tgvashworth:
A fighter's built-ins are materialised into their own assignments when they are hired. Nothing re-runs afterwards, so a built-in added to a profile later never reaches the fighters already hired from it. Those fighters silently keep the set they were hired with, and nothing in the app says the two have diverged.

## How it showed up

The n26 equip screen is gaining a tab per equipment list the fighter can buy from. A fighter reaches an equipment list by holding the `Collection` row among its built-ins, and an `Equipment List` column on the All Profiles sheet now attaches one during ingest.

An Ironhead Squats Charter Master showed only the Trading Post tab. The content was correct and the ingest had worked — the fighter was simply hired before the column existed:

```
PROFILE (Charter Master) built-ins:
   - Leader                          Subtype
   - XP                              Counter
   - Ironhead Squats Equipment List  Collection

DOUG'S OWN ASSIGNMENTS:
   - Leader                          Subtype
   - XP                              Counter
```

The profile has three; the fighter has two. Nothing is wrong with either row — there is just no path between them.

## Why it matters more now

Ingest can now **update** existing rows rather than skipping them, so a re-upload genuinely changes content that fighters were hired from. Before that, content was mostly write-once and the gap was easy to miss. Now every corrected sheet widens the divergence between what a profile says and what the fighters hired from it hold.

It also lands hardest while the sheets are still being written, which is exactly now: the first ingest of any new column reaches no existing fighter.

## The decision to make

Not obviously a bug to fix one way. Both readings are defensible and the choice should be deliberate:

- **A fighter is a snapshot.** What they were hired with is a historical fact, consistent with how the ledger treats everything else — a purchase records what was true at the time. Under this reading, propagation is wrong and the real gap is that there is no way to *see* the divergence or opt into repairing it.
- **A fighter reads their profile live.** Content fixes reach existing fighters, which is what an author correcting a sheet expects. Under this reading, some built-ins are structural (an equipment list is access, not kit) and should follow the profile, while others are genuinely bought things that must not be rewritten.

The second distinction may be the useful one: an equipment-list attachment is *access*, costs nothing, and adds no rating, so propagating it changes no money and no history. A built-in weapon is a thing the fighter owns, with a ledger entry behind it, and rewriting that retroactively would falsify the books.

## Worth considering

[...]

PR #2305, opened 2026-08-25
feat: a built-in added to a set reaches its existing holders within seconds

tgvashworth:
Part of #2165 (built-in propagation programme, chunk C4).

Until now, a built-in added to a profile's defaults only reached fighters hired *after* the change — everyone already hired stayed out of date forever. With this change, an authoring edit (form or spreadsheet ingest) that adds a member to a default-assignment set reaches every fighter already holding that set within seconds.

How it works:

- The authoring transaction files a durable `BuiltInPropagationTask` for the set (new model in `n26/core`, on the shared `gyrinx.state_machine` — forward-only PENDING → RUNNING → DONE | FAILED), then publishes a task message on commit. Filing is **append-only**: every edit inserts its own row, and a row is never reused — a standing row could otherwise be claimed and run against a library that predates the edit that found it, silently losing the change. A rolled-back edit files nothing. Redundant runs are no-ops, since the pass grants only what provenance shows is missing.
- The worker claims the row by state transition (a duplicate delivery loses the transition and stands down), resolves every holding gang — including option-set selectors via `ChosenProfileOption` — and reconciles each gang independently through `Operation.reconcile_defaults(carrier)`. Fighters that already have the built-in, or whose owner removed it, are left alone (provenance satisfaction, D3/D4). Archived gangs are skipped.
- A `*/5` scheduled sweep re-publishes messages for stale PENDING rows, marks abandoned RUNNING rows FAILED, and files a fresh row for a set whose latest run failed (stopping after 3 consecutive failures). This is the first task in production using `TaskRoute(schedule=…)` — **after deploy, verify the Cloud Scheduler job `…--gyrinx-scheduler--…sweep_built_in_propagations` exists and fires.**
- The gang's history records each run with a new `LedgerEvent.Kind.CAUGHT_UP`, no actor, folded per run: "Bruta gained Frag grenades — now part of what a Stimmer comes with".

**The running side ships behind a feature flag, off by default.** `propagate_built_ins` and the sweep both stand down while the `built-in-propagation` FeatureFlag is shut; filing is untouched, so a shut flag loses no work — rows accumulate PENDING and the first sweep after the flag opens drains the backlog. The flag is the kill switch: admin-toggleable, no redeploy. (Background work has no account to ask about, so `gyrinx.site.flags` gains `switched_on(slug)` — on for any availability but off.)

Go-live sequence: merge and deploy shut → verify the Cloud Scheduler job fires → create and open the `built-in-propagation` flag in the admin.

Removal-propagation is deliberately absent — that is chunk C6, an explicit previewed author action. Fighters hired before this ships are repaired by the C7 backfill.

Migrations: `n26.0029` (model + transition table + the new event kind). Chaos-tested through the manual task queue (rollback, duplicate delivery, dropped message, shared sets, option-set selectors, an edit landing while a run is in flight); browser-verified end-to-end on the dev server, including the history wording.

https://claude.ai/code/session_01491jPBkwQ8ok6UivFW3f2E


n26/tests/sandbox/test_builtin_propagation.py
```diff
@@ -0,0 +1,414 @@
...
+        assert (
+            "Ana gained Nerves of Steel — now part of what a Ganger comes with" in told
+        )
+        act = acts[
+            told.index(
+                "Ana gained Nerves of Steel — now part of what a Ganger comes with"
+            )
+        ]
+        assert act.actor == ""
+        assert act.subs == []
```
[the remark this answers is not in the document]
tgvashworth: Fixed in 0556e39c.

n26/maintenance.py
```diff
@@ -407,8 +408,17 @@ def delete_nameless_gang_type_view(request):
...
+#: ``n26.core.propagation``; their routes are declared here because this
+#: module is the edition's one door onto the task framework. The sweep
+#: is scheduled work: the framework provisions a Cloud Scheduler job
+#: from the declaration, and only there — the local backend fires no
+#: schedules, so dev and tests invoke the sweep function directly.
 task_routes = [
     TaskRoute(delete_nameless_gang_type, ack_deadline=600, min_retry_delay=60),
+    TaskRoute(propagate_built_ins),
+    TaskRoute(sweep_built_in_obligations, schedule="*/5 * * * *"),
 ]
```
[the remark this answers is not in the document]
tgvashworth: Fixed in 0556e39c.

n26/core/propagation.py
```diff
@@ -0,0 +1,287 @@
...
+            ).values_list("pk", flat=True)
+        )
+    ids.update(
+        Assignment.objects.filter(
+            chosen_options__default_set=default_set,
+            archived=False,
+            removes=False,
+        ).values_list("pk", flat=True)
+    )
+    return ids
```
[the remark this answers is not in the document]
tgvashworth: Fixed in 0556e39c.

n26/tests/sandbox/test_builtin_propagation.py
```diff
@@ -0,0 +1,414 @@
...
+    offer_option,
+    remove,
+)
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.fixture
+def player():
+    return User.objects.create_user("tom")
```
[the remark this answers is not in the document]
tgvashworth: Declined: n26 tests must not use repo-root conftest fixtures — n26/tests/CLAUDE.md forbids it (same-name fixtures shadow silently); shared n26 fixtures live in n26/tests/fixtures.py.

n26/tests/sandbox/test_builtin_propagation.py
```diff
@@ -0,0 +1,414 @@
...
+        profile = create_profile("Undecided", person_type, gang_type, price=50)
+        before = ReconcileObligation.objects.count()
+
+        with task_queue.capture():
+            with pytest.raises(RuntimeError):
+                with transaction.atomic():
+                    add_built_in(profile, create_rule("Second Thoughts"))
+                    raise RuntimeError("changed my mind")
+
+        assert ReconcileObligation.objects.count() == before
```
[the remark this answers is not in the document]
tgvashworth: False positive: the raise is inside pytest.raises, which swallows it — the statements after the block run on every test execution.

n26/tests/sandbox/test_builtin_propagation.py
```diff
@@ -0,0 +1,414 @@
...
+
+class TestTheDebtRidesTheEdit:
+    """Filing happens in the authoring transaction: an edit that commits
+    owes exactly one pass, and one that rolls back owes nothing."""
+
+    def test_a_rolled_back_edit_files_nothing_and_sends_nothing(
+        self, gang, person_type, gang_type, default_pack, task_queue
+    ):
+        profile = create_profile("Undecided", person_type, gang_type, price=50)
+        before = ReconcileObligation.objects.count()
```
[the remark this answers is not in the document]
tgvashworth: False positive: 'before' is read in the assertion after the pytest.raises block, which is reachable (the context manager swallows the raise).

n26/core/models/built_in_propagation.py
```diff
@@ -0,0 +1,71 @@
...
+whose publish can be lost, so the work is a durable row filed in the
+author's own transaction: however the message fares, the row says what
+is not yet applied, and a scheduled sweep re-publishes anything left
+standing (:mod:`n26.core.propagation`).
+
+One row is one pass, single-shot and strictly forward: PENDING →
+RUNNING → DONE or FAILED, no backward edges. Filing is append-only —
+every edit inserts its own row, and rows are never shared or reused.
+Reuse would race: a second edit attaching to a standing PENDING row can
+find that row claimed, and its library read, before the edit commits —
```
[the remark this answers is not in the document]
tgvashworth: Right — the description predated the redesign. Updated: filing is append-only, every edit inserts its own row, no PENDING-reuse constraint.

n26/tests/sandbox/test_builtin_propagation.py
```diff
@@ -0,0 +1,455 @@
...
+        profile = create_profile("Undecided", person_type, gang_type, price=50)
+        before = BuiltInPropagationTask.objects.count()
+
+        with task_queue.capture():
+            with pytest.raises(RuntimeError):
+                with transaction.atomic():
+                    add_built_in(profile, create_rule("Second Thoughts"))
+                    raise RuntimeError("changed my mind")
+
+        assert BuiltInPropagationTask.objects.count() == before
```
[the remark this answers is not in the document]
tgvashworth: False positive: the raise is inside pytest.raises, which swallows it — the statements after the block run on every execution.

n26/tests/sandbox/test_builtin_propagation.py
```diff
@@ -0,0 +1,455 @@
...
+class TestTheFilingRidesTheEdit:
+    """Filing happens in the authoring transaction, append-only: an
+    edit that commits gets a row and a pass of its own, and one that
+    rolls back files nothing."""
+
+    def test_a_rolled_back_edit_files_nothing_and_sends_nothing(
+        self, gang, person_type, gang_type, default_pack, task_queue
+    ):
+        profile = create_profile("Undecided", person_type, gang_type, price=50)
+        before = BuiltInPropagationTask.objects.count()
```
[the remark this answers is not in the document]
tgvashworth: False positive: 'before' is read in the assertion after the pytest.raises block, which is reachable.

n26/core/models/built_in_propagation.py
```diff
@@ -0,0 +1,71 @@
+"""The propagation a set change files against everything already holding it.
```
tgvashworth: This language is horrible.

"Records a change to a built-in set and tracks async propagation"

n26/core/models/ledger.py
```diff
@@ -104,6 +104,12 @@ class Kind(models.TextChoices):
         PURCHASED = "purchased", "Purchased"
         ADDED = "added", "Added"
         GRANTED = "granted", "Granted"
+        # A built-in that arrived after the fact: an author added it to
+        # what the thing comes with, and a background pass granted it to
+        # everything already holding the thing. Its own kind, because a
+        # reader asking "why did this appear on my card?" is owed the
+        # answer, and the plain ``granted`` cannot give it.
```
tgvashworth: Too long -- just refer to built-in propagation

n26/core/propagation.py
```diff
@@ -0,0 +1,298 @@
+"""Propagation — a set change reaches everything already holding it.
```
tgvashworth: a "built-in set change"... and mention async

n26/core/propagation.py
```diff
@@ -0,0 +1,298 @@
...
+passes this buys are no-ops.
+
+The claim itself is the row's PENDING → RUNNING transition, validated
+under the row's own lock — of two concurrent deliveries one wins and
+the other stands down silently, a routine outcome. Publishing is
+fire-and-forget and can lose a message, so a scheduled sweep
+re-publishes rows left PENDING, declares a pass whose worker died
+lost, and files a fresh row after a failure. Retry is always a fresh
+row: the graph is strictly forward, and an ended row is a record.
+"""
```
tgvashworth, line 35: This is significantly duplicated vs the model, and we only need to keep one. I suggest reducing the model comment, and then reviewing this to improve clarity and simplicity.

n26/library/authoring.py
```diff
@@ -1097,6 +1097,14 @@ def add_default_member(
...
         member.clean()
     member.save()
+    # Everything already holding this set is now owed the new member.
+    # The filing rides this transaction — made with the edit, rolled
+    # back with it — and a background pass applies it after commit.
+    # Every path that adds a member comes through here, so this is the
+    # one place the pass is filed.
+    from n26.core.propagation import file_propagation_task
+
+    file_propagation_task(default_set)
```
tgvashworth, line 1111: Does this need better error handling? If the create fails... the built-in has still been added

n26/library/authoring.py
```diff
@@ -1097,6 +1097,14 @@ def add_default_member(
         # cannot say either.
         member.clean()
     member.save()
+    # Everything already holding this set is now owed the new member.
+    # The filing rides this transaction — made with the edit, rolled
+    # back with it — and a background pass applies it after commit.
+    # Every path that adds a member comes through here, so this is the
+    # one place the pass is filed.
+    from n26.core.propagation import file_propagation_task
```
tgvashworth, line 1109: probably should move this import to top of file

n26/core/models/built_in_propagation.py
tgvashworth: Rewritten in 936c6e64 with your wording as the opener — the model docstring is now 7 lines; the race story lives only in n26/core/propagation.py.

n26/core/models/ledger.py
tgvashworth: Shortened in 936c6e64 — two lines referring to built-in propagation.

n26/core/propagation.py
tgvashworth: Reworded in 936c6e64 — opens with 'a built-in set change … asynchronously'.

n26/core/propagation.py
tgvashworth: Deduplicated in 936c6e64: the model docstring is reduced to a stub pointing here, and this one was tightened (append-only paragraph cut to four sentences).

n26/library/authoring.py
tgvashworth: Good catch — real for a caller holding no transaction: the member would commit and the filing be lost. Fixed in 936c6e64 by making add_default_member @transaction.atomic, so the member and its filing commit or roll back together regardless of the caller.

n26/library/authoring.py
tgvashworth: Declined: library imports core only inside function bodies — the n26/CLAUDE.md boundary rule (module-level imports of core from library break the one-way import discipline and can break boot). Happy to revisit the rule itself if you want, but this call site follows it.

PR #2343, opened 2026-08-28
feat: convert Chaos God onto slots and picks

tgvashworth:
The Chaos God choice now converts onto slots and picks in one maintenance-console run, both doors at once. Variants are not in this operation.

Until now that question lived as two `Affiliation` offers labelled Chaos God: one on the Hidden built into Chaos Helot Cult, one on the Chaos Corrupted affiliation (the Variant pick). After this run they are two slot rows of one Chaos God type, drawing on the same Chaos Gods picklist; the four gods (Architect of Fate, Blood God, Dark Prince, Plague Lord) become pickables with nothing moved — they have no payload; every stored god pick — live and archived — is rewritten as a pick on the same assignment. A spread of gangs must still read the same, and every reached live gang must still reconcile, or the write unwinds.

Chaos Corrupted stays an Affiliation. The Affiliation model and column stay. Fossils and other affiliation-labelled offers (already-converted Outcast, Variants) are left standing. `min_picks=0` on both slots, so an unanswered Helot is not nagged.

`CONVERT_CHAOS_GOD` uses advisory lock `826_020_610`. It does not share `REPAIR_DOUBLED_REFUNDS`'s `826_020_609`.

History wording moves for converted god picks: `_kindword` reads the slot type, so a past pick that said “affiliation” now says “chaos god” (72 production picks). Captures do not see the history page. That is the same trade-off as the Outcast conversion (#2334), covered by `test_a_god_pick_is_reworded_to_chaos_god`.

Production (read-only plan, 2026-08-28): 59 live god picks, 13 archived, 84 live gangs reached. Helots mostly unanswered; Chaos Corrupted mostly answered. Both doors labelled Chaos God, same menu, every live pick gang-hosted. No Chaos God slot type yet. The Helot hidden is named Chaos God (not a longer label); the default fixture matches that name. The plan is OK; this PR does not apply it.

## How to try it

1. Deploy, then open the maintenance console as a superuser.
2. Run **n26: the Chaos Gods become picks**.
3. Preview lists the creates, the two offer-to-grant swaps (Helot hidden and Chaos Corrupted affiliation — not a pickable), and every pick rewrite. Apply is one transaction. It does not enqueue `audit_reconcile`; run that by hand afterwards.
4. Check an unanswered Helot gang (no “Chaos God — 0 of 1 chosen”), a Helot that picked a god, a Chaos Corrupted gang that never picked a god, one that did, and one that re-chose: the sheets should match what they said before.
5. Re-choosing a god on the converted machinery should still work on both doors. History names a converted pick a “chaos god”.

A second run is a no-op (`nothing to convert`). Do not run this against a database that already has a Chaos God slot type of that name in the default pack.

Variants (including None) stay as a follow-up operation. Do not drop `OFFERABLE_KINDS` or the Assignment column in this PR.

## Tests

`pytest n26/tests/sandbox/test_conversion_chaos_god.py n26/tests/test_maintenance_console.py n26/tests/sandbox/test_conversion_affiliation.py n26/tests/sandbox/test_gang_books.py` — 98 passed. `test_gang_books.py` still builds Chaos corruption as an Affiliation until Variants.

<div><a href="https://cursor.com/agents/bc-c6f07668-ade5-439b-8e88-8204263594b7?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c6f07668-ade5-439b-8e88-8204263594b7&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


tgvashworth:
Reviewed and tested this against production data and at production volume. **One blocker, otherwise it looks solid.**

## Blocker: duplicate advisory lock key

```python
LOCK_KEYS = {
    ...
    Operation.CONVERT_CHAOS_GOD: 826_020_609,
    Operation.REPAIR_DOUBLED_REFUNDS: 826_020_609,   # already on main
}
```

`REPAIR_DOUBLED_REFUNDS` holds `826_020_609` on `main` already, and it is not a retired slug — it was added the same day and has a live view. So two runnable console operations now share one lock.

The note directly above `LOCK_KEYS` describes this exact failure: the view's running-guard only sees its own operation's records, so one repair can enqueue while the other holds the lock, stand down without writing, acknowledge its message, and **leave a record saying RUNNING for ever**. Two brand-new one-off repairs are likely to be run in the same sitting, so this isn't theoretical.

`826_020_610` is free. Worth adding a test that `LOCK_KEYS` values are unique — nothing currently catches this, which is how it got here.

## Verified against production (read-only)

`plan_chaos_god()` is clean — no problems — and finds the system by structure exactly as intended:

- **84 live gangs reached** (117 gangs hold a door; the other 33 are archived, rewritten but not held — correct)
- **82 steps**: 1 slot type, 4 pickables (no payload to move), 1 picklist, 2 slots, 2 swaps, **72 rewrites** (35 Helot door / 37 Corrupted door), 0 spares
- both slots derive `label` = `Chaos God` from their offers, and get distinct qualifiers `Hidden` / `Chaos Corrupted`
- both offers name the same menu; the Variant, Outcast and fossil offers are left alone

Also checked the edge case the two same-named slots create: both slots carry the same `kind_label`, so a gang holding both doors would show two identically-labelled questions and `Computed.choice()` would return whichever came first. **No production gang holds both doors** (0), so it doesn't arise today — but it's worth knowing it's latent if Chaos Corrupted ever becomes available to Helots.

## Verified at production volume locally

Built an 84-gang population on the real content mirror in the production shape (both doors, gods picked and unpicked, re-choices leaving archived picks, archived gangs), then ran the real apply:

- **6.8s**
- captured **all 87 gangs** before and after myself, from outside the change — **zero differences**
- **zero reconcile failures**
- second run reports `nothing_here` — a clean no-op

`pytest n26`: 4222 passed, 1 xfailed.

## Two notes, neither blocking

**The fixture diverges from production in the one place that matters.** The default `prod_shape` names the Helot hidden `Chaos God — Helots`, but production names it `Chaos God` — which routes through `_door_qualifier`'s fallback and yields qualifier `Hidden` instead. That production path only has a plan-level test (`test_a_hidden_that_shares_the_slot_name_is_told_apart_by_kind`); no apply, no capture. I exercised it with a full apply against the mirror and it is fine, but this is the same shape of gap that hid the label bug in #2334 — the fixture agreeing with the code rather than with production. Suggest making the default fixture use production's name and keeping the current one as the variant.

**History wording moves for god picks.** `_kindword` reads the slot type, so a past god pick reads "chaos god" where it read "affiliation" — 72 picks. Deliberate and tested (`test_a_god_pick_is_reworded_to_chaos_god`), and the same trade-off accepted in #2334, but it sits against the rule stated in `conversion/base.py` and isn't mentioned in the description.

Minor: `chaos_god.py` imports `_affiliations_on` and `_qualifier_for` from `affiliation.py`. They're shared helpers now rather than that module's privates — `base.py` may be the better home.

PR #2347, opened 2026-08-28
feat: convert Variant onto slots and picks

tgvashworth:
The shared house-gang Variant question — Chaos Corrupted, Genestealer Cult Corrupted, Malstrain Corrupted, or None — moves onto an optional Variant slot and picks. After deploy, a superuser runs this once from the maintenance console, then `audit_reconcile` by hand.

## What changed

House gangs are asked Variant as a slot (`min_picks=0`, `assigned_to=gang`), granted from the same shared modifier the seven gang types already carry. The three corruptions become pickables and keep their payloads, including Chaos Corrupted's Chaos God grant, so un-choosing Variant still retracts the god.

**Stored None picks are archived, and those cards will change.** Today they print the word “None”. After the conversion they print nothing: an optional slot with nothing chosen. The picker already has a None row. That is a deliberate product call, not an accident the proof papered over. `unanswered_as` is what lets capture treat printed None and unanswered as equal for that question only; apply then checks every planned `ArchivePick` landed by identity, because that proof can no longer see those archives.

Affiliation stays. Outcast and Chaos God are not converted again. The vestigial hidden “Variant” is a carrier of that one swap, not a second door.

## How to try it

1. Maintenance console → **n26: the Variants become picks**. GET shows the plan (creates, shared swap, rewrites, how many Nones it archives) and states that cards printing None will print nothing after. POST applies.
2. A house gang on None: the sheet asks Variant with nothing chosen, and does not nag “0 of 1”.
3. A Chaos Corrupted gang is still asked its Chaos God. Un-choose Variant and the god goes with it.
4. Then run the existing `audit_reconcile` operation. Do not nest it on this Backfill.

Retry after a successful flip is `nothing_here`. A standing Variant slot type is a refusal, not a skip.

## Production plan and timing

`plan_variant()` is ok against production. Carriers are exactly the seven house gang types plus the hidden “Variant” (0 live assignments). No Variant slot type yet. Chaos Corrupted already grants the Chaos God slot.

Counts: 189 live None, 19 archived None, 69 live corruptions (36 Chaos / 15 GSC / 18 Malstrain), 34 archived corruptions. Preview archives 208 Nones and rewrites 103 corruption picks. It reaches 1022 live house gangs.

A local rebuild at production’s pick counts, after Chaos God, applied in 68.4s at 39k assignments; reconciling all 1022 gangs is the dominant cost. Production is denser, so expect roughly 2 minutes — comfortable against the 600s ack deadline. The run holds a write lock on those 1022 gangs for that window, so it is better at a quiet time than at peak.

A live Van Saar on None already shows Variant only on the gang sheet (nine models, none of them asked). `gang_alone` is the right grant reach.

Independent capture of all 1022 gangs, without this conversion’s canonicalisation: exactly 208 differ, and only as `('Variant', 'None')` → `('Variant', '')`. 0 reconcile failures. A second run reports `nothing_here`.

<div><a href="https://cursor.com/agents/bc-f2894c7a-41e2-407a-b223-ac90a2648c91?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f2894c7a-41e2-407a-b223-ac90a2648c91&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/n26/library/conversion/__init__.py
+++ b/n26/library/conversion/__init__.py
@@ -7,10 +7,12 @@ Each system's ``plan_*`` reads the database and returns a frozen Plan.
 from n26.library.conversion.affiliation import plan_outcast_affiliation
 from n26.library.conversion.base import ConversionRefused, Plan, apply
 from n26.library.conversion.chaos_god import plan_chaos_god
+from n26.library.conversion.variant import plan_variant
 
 SYSTEMS = {
     "outcast_affiliation": plan_outcast_affiliation,
     "chaos_god": plan_chaos_god,
+    "variant": plan_variant,
 }
 
 __all__ = [
@@ -20,4 +22,5 @@ __all__ = [
     "apply",
     "plan_chaos_god",
     "plan_outcast_affiliation",
+    "plan_variant",
 ]
--- a/n26/library/conversion/base.py
+++ b/n26/library/conversion/base.py
@@ -454,6 +454,28 @@ class SwapSharedCarrier:
 
 
 @dataclass(frozen=True)
+class ArchivePick:
+    """A stored pick archived in place — the declared ledger change for
+    a printed “None” that an optional slot already says with nothing
+    chosen. Already-archived picks stay archived; this does not revive
+    them. No money, no event: the conversion exception."""
+
+    assignment_id: object
+    gang: str
+    name: str = "None"
+
+    def say(self):
+        return f"archive pick {self.assignment_id} ({self.gang}) — “{self.name}”"
+
+    def perform(self, made):
+        from n26.core.models import Assignment
+
+        pick = Assignment.objects.get(pk=self.assignment_id)
+        if not pick.archived:
+            pick.archive()
+
+
+@dataclass(frozen=True)
 class RewritePick:
     """One stored choice, re-said as a pick: the old kind's column moves
     to ``pickable``, the anchor it already hangs from (``caused_by``)
@@ -532,6 +554,11 @@ class Plan:
     #: True when the system simply is not here — nothing to convert and
     #: nothing wrong: the apply is a clean no-op.
     nothing_here: bool = False
+    #: Choice pairs this conversion treats as unanswered after the write.
+    #: Archiving a printed “None” leaves the same question with nothing
+    #: settled; capture otherwise sees the stored name become "". Each
+    #: pair is ``(kind_label, chosen)``.
+    unanswered_as: tuple = ()
 
     @property
     def ok(self):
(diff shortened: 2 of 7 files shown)


tgvashworth:
Reviewed and tested this against production data and rebuilt production's scale locally. **No blocker. One thing that needs a decision rather than a fix, and one cheap hardening.**

Both points I raised on #2343 landed: lock keys are unique (`CONVERT_VARIANT: 826_020_611`) and `test_its_lock_is_not_shared` now asserts it.

## This one is 13× bigger than the previous two — but it fits

The Variant offer sits on seven house gang types, so production's reach is much larger than the earlier conversions:

| | Outcast | Chaos God | **Variant** |
|---|---|---|---|
| gangs reached | 76 | 84 | **1,022** |
| assignments in them | 4,202 | — | **66,542** |
| steps | 164 | 83 | **318** |

That mattered because the whole thing is one transaction that `SELECT FOR UPDATE`s every reached gang and reconciles all of them before committing, and the route keeps `ack_deadline=600` against a 600s Cloud Run timeout — with `MAX_ATTEMPTS=2`, a run that overran would fail twice and end as FAILED having written nothing.

So I measured it rather than guessed. Built 1,022 gangs on the real content mirror at production's exact pick counts (103 corruptions, 208 Nones), ran the Chaos God conversion first to match production's current state, then applied:

- **apply: 68.4s** at 39,317 assignments
- reconciling all 1,022 gangs is the dominant cost (~65s of it)
- production carries 1.7× my assignment density, so expect **roughly 2 minutes**

Comfortable against the 600s deadline — about a 5× margin. **Not a blocker.** Worth knowing operationally that it holds a write lock on 1,022 gangs for those ~2 minutes, so it's better run at a quiet time than at peak.

## The 208 "None" gangs do genuinely change what the card says

This is the one I'd want stated explicitly in the description before it runs.

I captured all 1,022 gangs before and after myself, *without* the conversion's own canonicalisation. Exactly **208 gangs differ, and nothing else does**:

```
choices: [('Variant', 'None')] -> [('Variant', '')]
```

Those cards print the word "None" today and will print nothing. `unanswered_as` is what lets this pass the proof, and the reasoning in `ArchivePick`'s docstring is sound — an optional slot with nothing chosen says the same thing to a player as a printed "None". But `conversion/base.py` opens by promising that a conversion which would change what a reader is told never lands, and this is the first declared exception to that. 208 gangs is not nothing. It reads to me as a defensible product call, just one that should be made deliberately and written down rather than discovered later.

## The proof no longer watches the thing this conversion newly does

`_canonicalize_unanswered` is applied to **both** sides of the comparison. That's what makes it work — but it also means a `("Variant","None")` that was *not* archived compares equal to one that was. For the 208 archives, which are this conversion's new behaviour, the capture can no longer tell success from no-op.

I checked it independently: **208 of 208 archived**, so the mechanism is correct. This is defence-in-depth, not a defect. But it'd be cheap to assert after the write that every planned `ArchivePick` is now archived, so the one thing the proof stopped watching is still checked by something.

## Everything else verified

- plan against production: clean, no problems, `unanswered_as` correctly scoped to the single `('Variant', 'None')` pair
- 3 pickables, 1 picklist, 1 slot, **1 shared grant** — `SwapSharedCarrier`'s first real use, and the offer's factoring is preserved as one row
- slot label derived from the offer (`Variant`), not hardcoded
- the guard refusing to run while a corruption still offers a Chaos God choice is the right ordering discipline
- Chaos Corrupted stays an Affiliation and keeps its Chaos God grant through the conversion; the chain still retracts
- after the apply: **0 reconcile failures across all 1,022 gangs**, and a second run reports `nothing_here`
- `pytest n26`: 4280 passed, 1 xfailed

Minor, as on #2343: `variant.py` reaches into `affiliation.py` for `_affiliations_on`, `_offers_of_kind` and `_qualifier_for`. Three conversions now share them — they look like `base.py` residents.

Issue #2296, opened 2026-08-23
feat: add n26 Ranks, advancement picks and maximum characteristics


tgvashworth:
Part of the **N26 Campaigns** milestone. Stream B — not behind the feature flag. Consumes the roll metadata added by the lasting-effects ticket.

## Current state

`Miniature.xp` and `Miniature.xp_target` exist, and `xp_target`'s help text already says it *"should come from the Model Ranks table once ranks are modelled"*. Nothing else does.

## What to build

- **Ranks** — a rank table (threshold → rank), driving current rank, next threshold, and how many advancements are owed.
- **Advancement tokens** — one per rank gained, spent on a pick. XP adjusted down must notice and prompt. Removing an advancement offers a refund. **Users never see the word "token".**
- **Maximum characteristics** — `Stat` (`n26/library/models/statline.py:62`) has no maximum today. ⚠️ **Watch `is_inverted`**: a 2+ beats a 3+, so "at maximum" runs in opposite directions per stat and a naive comparison will be wrong for half the statline. Warn on an illegal pick but allow proceeding, per the design canvas.

Advancements raise Credits Value. `op.learn` (`n26/core/operations.py:1350`) is already the exact pattern — a free assignment that still carries a rating.

## Fits slots and picks

An advancement is a slot; table results are pickables. **Shares the roll-metadata dependency** with lasting effects, so that ticket lands the fields and this one consumes them.

⚠️ One caveat: the two tables have different semantics. The lasting-effect D66 is a **band lookup**; the advancement 2D6 is a **threshold** ("select any result you rolled high enough for"). Check that one field design serves both.

## Do not port n23's advancement flow

`n23/core/views/fighter/advancements.py` is 1360 LOC and a five-step wizard. The design canvas is explicit that N26's system is dramatically simpler.

## Content

Ship the advancement table and the rank thresholds as content in this ticket — a ticket is not done if nobody can use what it built.

---

**Consumes:** #2293 (roll metadata). Check the band-vs-threshold caveat against it before it commits to a field design.

PR #2492, opened 2026-09-06
fix: stop Lasting Injuries taking a Stat past its minimum or maximum

tgvashworth:
## Summary

- In n26, stacked Lasting Injuries pushed a characteristic past the rulebook's floor: a fighter with three Spinal Injuries showed **S 0**, a fourth would show **-1**, and a Save worsened three times read **9+**. The book says a characteristic is never reduced below its minimum or improved past its maximum, and the part of an injury that would do so is disregarded rather than carried over.
- `Stat` now carries `minimum` and `maximum` (library migration `0093`). They are stored as the number the cell stops at, so a roll-target stat's minimum is the *larger* number: Save runs 6 (6+) to 3 (3+). The data migration fills the thirteen model characteristics from the rulebook table in every pack and leaves weapon-only stats (SR, LR, AP, L) blank. **No content authoring is needed**; the authoring form and `create_stat` take the two fields for future stats.
- `apply_changes` (`n26/core/render.py`) folds changes **one at a time in the order they were acquired**, stopping at the limits after each. That is what makes "the surplus is not kept" true: ten Spinal Injuries then a Bionic Arm leaves Strength at 2, not 1. A value an owner has already set past a limit is left where it is rather than pulled back. The base value and `set`-mode changes are not clamped, only what folds on top of them.
- The card's tooltip now says why an injury shows no further effect: `S changed by Spinal Injury (lasting injury), Spinal Injury (lasting injury), Spinal Injury (lasting injury). Cannot get any worse.`

Strength is one shared `Stat` row for the Model and Weapon shapes, so a weapon's Str also stops at 1 and 10.

## Test plan

- [x] `pytest n26 -n 4`: 5733 passed, 1 xfailed
- [x] New colocated tests `n26/core/test_render.py` cover the floor, the ceiling on inverted stats, surplus discarded then an improvement, acquisition ordering, values already past a limit, unlimited stats, non-numeric values and `set`
- [x] `n26/library/test_statline.py` covers `Stat.bounds()` / `Stat.shift()`
- [x] `n26/tests/sandbox/test_lasting_effects.py`: three Spinal Injuries from the shipped tables leave S at 1 with every injury listed; five injuries then a bionic give 2
- [x] Migration smoke-tested on a fork of `gyrinx_main`: 13 rows filled, 4 weapon stats left blank, reverse migration drops the columns cleanly; `makemigrations --check` reports no changes
- [x] Tooltip rendered through the cotton component for both the minimum and maximum case
- [ ] After deploy: open a model with stacked injuries on `/n26/` and check the held cell and its tooltip

Refs #2296 (lands the `Stat.maximum` field that ticket calls for; the advancement flow itself is not here)

https://claude.ai/code/session_014tNMFpfFYuiGgafAwBGB4J


tgvashworth:
Review pass: sorted the change list once (Copilot), cleared a stale held_at after a later improvement (CodeRabbit), moved acquired onto Node, held_note in Python, IntegerField + Stat.clean(), seed fills blank limits on existing rows. The min/max CheckConstraint suggestion is acknowledged, not taken: a reversed pair now fails clean() in words, and a stored one fails safe (nothing moves). Fixed in 9f412ed8d.

